PR #1, opened 2015-03-10
First tokenizer steps

whatyouhide:
I implemented a first (poor) draft of the tokenizer which only supports tokenizing strings and the `msgid` and `msgstr` keywords. I'm opening a PR this soon so that I know if I'm on the right track.

@josevalim and I discussed tokenizing and parsing on IRC. We were thinking about whether to roll out both a handwritten tokenizer and a handwritten parser or to write a tokenizer and than using with `yecc` (like Elixir).

As far as I looked into `.po` files, the grammar seems very simple. This means that we may be able to use a simple handwritten parser. Let's hear what everybody else has to say :).

### DIFF
--- a/lib/gettext/po/tokenizer.ex
+++ b/lib/gettext/po/tokenizer.ex
@@ -42,13 +42,25 @@ defmodule Gettext.PO.Tokenizer do
   @spec tokenize_line(binary, pos_integer, [token]) :: [token]
   defp tokenize_line(str, line, acc)
 
+  # End of file.
+  defp tokenize_line(<<>>, _line, acc) do
+    Enum.reverse acc
+  end
+
   # Go to the next line.
   defp tokenize_line(<<?\n, rest :: binary>>, line, acc) do
     tokenize_line(rest, line + 1, acc)
   end
 
   # Skip whitespace.
-  defp tokenize_line(<<char, rest :: binary>>, line, acc) when char in @whitespace_no_nl do
+  defp tokenize_line(<<char, rest :: binary>>, line, acc)
+      when char in @whitespace_no_nl do
+    tokenize_line(rest, line, acc)
+  end
+
+  # Comments.
+  defp tokenize_line(<<?#, rest :: binary>>, line, acc) do
+    {_comment_contents, rest} = to_eol_or_eof(rest, "")
     tokenize_line(rest, line, acc)
   end
 
@@ -67,13 +79,8 @@ defmodule Gettext.PO.Tokenizer do
 
   # String start.
   defp tokenize_line(<<?", rest :: binary>>, line, acc) do
-    {str, rest} = tokenize_string(rest, line, "")
-    tokenize_line(rest, line, [{:string, line, str}|acc])
-  end
-
-  # End of file.
-  defp tokenize_line(<<>>, _line, acc) do
-    Enum.reverse acc
+    {token, rest} = tokenize_string(rest, line, "")
+    tokenize_line(rest, line, [token|acc])
   end
 
   # Parses the double-quotes-delimited string `str` into a single `{:string,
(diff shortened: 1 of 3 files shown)


lib/gettext/po/tokenizer.ex
```diff
@@ -0,0 +1,104 @@
...
+  # (since that was needed to identify the start of a string). Returns a tuple
+  # with the contents of the string and the rest of the original `str` (note
+  # that the rest of the original string doesn't include the closing double
+  # quote).
+  @spec tokenize_string(binary, pos_integer, binary) :: {token, binary}
+  defp tokenize_string(str, line, acc)
+
+  defp tokenize_string(<<?", rest :: binary>>, _line, acc),
+    do: {acc, rest}
+  defp tokenize_string(<<?\\, ?n, rest :: binary>>, line, acc),
```
josevalim: Alternatively, you can write:

    defp tokenize_string(<<?\\, char, rest :: binary>>, line, acc) when char in @escapeble_chars do ...

And then:

    defp escape_char(?n), do: ?\n
    defp escape_char(?\\), do: ?\\
    ...

Another question is: what happens when you see `\\` but not followed by a escapable char? Like `\a`? Syntax error?

josevalim:
Beautiful!

lib/gettext/po/tokenizer.ex
whatyouhide: The problem with the `char in @escapable_chars` approach (which I tried first) is that you'd have to map chars to other chars. What I mean is: `<<?\\, char, _>>` binds `char` to, e.g., `n` or `"` but you want different behaviours for those two cases, turning `<<?\\, ?n>>` into `?\n` and `<<?\\, ?">>` into `?"`. I couldn't come up with a nicer way to do it.

josevalim: That's exactly what the escape_char above would do. :)

josevalim:
```elixir
defp tokenize_string(<<?\\, char, rest :: binary>>, line, acc) when char in @escapable_chars,
  do: tokenize_string(rest, line, <<acc :: binary, escape_char(char)>>)

defp escape_char(?n), do: ?\n
defp escape_char(?t), do: ?\t
defp escape_char(?"), do: ?"
defp escape_char(?\\), do: ?\\
```

whatyouhide: Jesus I completely skipped half of your comment. I can't believe it, I just didn't see it :eyeglasses: 

**PS** I just found [this](http://stackoverflow.com/questions/10218631/printing-with-gettext) StackOverflow answer which suggests to look at the source code for `gettext`'s own `.po` parser to find out the supported escape codes. Apparently, a lot more are supported! :)

josevalim: You know what, let's keep it simple. Let's ignore the other escape codes and let's be sure to raise if you use `\` not followed by a escapable char.

whatyouhide:
Hey @josevalim, I added basic support for comments (I'm still not sure if we should tokenize things like tags in comments, we'll see later on). Going to work on parsing pluralizations soon!

josevalim:
Let's not worry about pluralization nor contexts for now. Let's ship v0.1 and then worry about additions. For example, some gettext implementations do not provide contexts, maybe we don't need them. Let's get something that works out and then we add features as we need them. :)

whatyouhide:
Ok, perfect! Should we start merging this PR then, so I'll move to parsing?

josevalim:
Sure, let's remove contexts and merge this. :) Also, feel free to merge it yourself.

About the comments, the tokenizer may even completely ignore them and not feed them to the parser. I particularly don't see we needing them down the road.

josevalim:
(At least not for now)

whatyouhide:
@josevalim for now they're pretty much useless, but many gettext tools rely on them in order to, e.g., know where the original string is in the source code or similar stuff.

What about this: I'll leave the comment parsing but only as something like "hey, if I see a `#` I'll just ignore the entire line up to - but not including - the final `\n`". What do you think?

I'm also going to remove parsing `msgctxt`, sorry I missed that :).

(**PS** I don't see the "Merge" button at the bottom of this page :( )

josevalim:
@whatyouhide right, we may need them later on, when merging stuff and so on. Fair enough. You should see the merge button now, my bad. :)

josevalim:
:heart: Let me know if you want any help when tackling the parser. :)

PR #3, opened 2015-03-18
.po parser (yecc-based)

whatyouhide:
Sorry it took me so long but I've been super busy and I've also been fighting with a nasty bug (read the commit messages if you want to get as angry as I did :smile:).

The syntax errors should be refined a little but the core functionality is there.
Let me know what you think for now!

### DIFF
--- a/lib/gettext/po/parser.ex
+++ b/lib/gettext/po/parser.ex
@@ -1,0 +1,28 @@
+defmodule Gettext.PO.Parser do
+  @moduledoc false
+
+  defmodule Translation do
+    @moduledoc false
+    defstruct [:msgid, :msgstr]
+  end
+
+  @doc """
+  Parses a list of tokens into a list of translations.
+  """
+  def parse(tokens) do
+    case :gettext_po_parser.parse(tokens) do
+      {:ok, translations} -> Enum.map(translations, &convert_to_struct/1)
+      {:error, _}         -> raise "syntax error"
+    end
+  end
+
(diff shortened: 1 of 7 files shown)


lib/gettext/po/parser.ex
```diff
@@ -0,0 +1,28 @@
+defmodule Gettext.PO.Parser do
+  @moduledoc false
+
+  defmodule Translation do
```
josevalim: I would move this up to `Gettext.PO.Translation` and document it. That's because it is likely folks will be using them directly. On the other, you can wait until it happens.

josevalim:
This looks great! Feel free to merge this. <3 

The next step is a function in `Gettext.PO` that parses a given file. If we are going with elixir naming, it should likely be called `parse_file` (and maybe followed by a `parse_string` too). However, if we are going to use gettext naming, we may want to search if there is a command name we could [steal from here](https://www.gnu.org/software/gettext/manual/html_node/index.html#SEC_Contents) (chapter 9 in particular).

whatyouhide:
@josevalim moved `Gettext.PO.Parser.Translation` to `Gettext.PO.Translation` (way better, thanks!) and merged :).

whatyouhide:
@josevalim with regards to the naming of the parsing functions, I think we should stick to Elixir conventions right now. Parsing a `.po` file isn't really something that gettext does "in public", I think it's more like an internal process and I think it should be an internal process in this library as well.

What do you think? I'm going to start working on the implementation of those function (regardless of what their name will be) in the meantime.

josevalim:
@whatyouhide perfect! :tada:

PR #4, opened 2015-03-21
Errors not exceptions

whatyouhide:
I think I could've pushed this right away instead of opening a PR, but a second opinion is always extremely helpful.

I just removed all the `raise` statements from both the parser and the tokenizer, making them return `{:ok, stuff} | {:error, line, reason}` tuples instead. I kind of stole this from EEx.

This way, we can control errors better: for example, when parsing a file instead of a string we can report the line **and the file** where the error occurred instead of the line only. Raising in the tokenizer/parser makes code a little bit harder to control, since every function call could fail badly and that may not always be what we want.

If you're ok with this, I'll merge :). Thanks!

### DIFF
--- a/lib/gettext/po/exceptions.ex
+++ b/lib/gettext/po/exceptions.ex
@@ -9,15 +9,3 @@ defmodule Gettext.PO.SyntaxError do
     %__MODULE__{message: msg}
   end
 end
-
-defmodule Gettext.PO.TokenMissingError do
-  defexception [:message]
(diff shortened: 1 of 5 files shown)


josevalim:
:+1:. We can provide `parse_file!` variants later in `Gettext.PO` in order to provide the raising versions.

whatyouhide:
Exactly. Merging then, thanks!

PR #5, opened 2015-04-01
`parse_string/1` and `parse_file/1`

whatyouhide:
Hey there @josevalim,

I implemented `Gettext.PO.parse_string/1` and `Gettext.PO.parse_file/1`. I also got carried away and implemented bang variants for both those functions, and tweaked `Gettext.PO.SyntaxError` a little bit in order to be more consistent with Elixir syntax error formatting (`file:line: reason`).

I'm not 100% sure we'll need the bang variants, but I'm willing to nuke them if you don't think they'll be useful.

I would also love your opinion on another thing: errors. As of now, I'm always returning `{:error, line, reason}` tuples instead of formatting the line and the reason into something more easy to read. Is that ok?

As always, thanks a lot :heart:!


josevalim:

josevalim:
:shipit:

PR #7, opened 2015-04-06
Add a .travis.yml file

whatyouhide:
Hey @josevalim, what do you think about adding Travis support to gettext? I know it's a bit soon but what the heck, we're probably adding it at some point anyways :).

If you agree, you would have to add the repo to Travis since I don't have admin privileges here.

I also added my email to the recipients if that's ok; if not, just let me know and I'll leave only yours. Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+language:
+  elixir
+elixir:
+  - 1.0.0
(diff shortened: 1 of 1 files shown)


josevalim:

PR #8, opened 2015-04-13
Pluralisation

whatyouhide:
Hey there,

I've been working on support for pluralization on the last couple of days and this is the result. I'd very much appreciate some feedback :smiley:. Consider this is the first iteration, there is still a lot of polishing to do but the core principles should be there.

Let me know!

Just for reference, the important commits are:

* 43f0c49 (tokenizing)
* ae198a9 and c0c8262 (parsing)
* df676f0 and 0d89075 (`Pluralizer` behaviour and default pluralizer)
* 0b57856 (`lngettext/5`)

The rest are mostly refactoring, minor fixes and docs!


lib/gettext/plural.ex
```diff
@@ -0,0 +1,415 @@
+defmodule Gettext.Plural.Helpers do
+  @moduledoc false
+
+  @doc """
+  Tells if the last digit of `n` is one of `digits`.
+
+  Can be used in function guards.
+  """
+  defmacro ends_in(n, digits) do
```
josevalim: Why not a private macro in Gettext.Plural? It is fine as long as it is at the top!

whatyouhide: Whoops, my bad. I thought you could only use private macros inside functions, turns out you can use them in guards too :).

lib/gettext/po/tokenizer.ex
```diff
@@ -77,10 +85,34 @@ defmodule Gettext.PO.Tokenizer do
     end
   end
 
-  # String start.
+  # `msgstr`.
+  defp tokenize_line("msgstr[" <> <<rest :: binary>>, line, acc) do
```
josevalim: In theory, those will be tokenized as three separate thing. You would have a token for `[`, another for integers and another for `]`. The grammar would put all four entities together.

In practice though, what you did is ok.

lib/gettext/po/translation.ex
```diff
@@ -2,17 +2,21 @@ defmodule Gettext.PO.Translation do
...
+    * `msgstr` - the translated string if there's no pluralization, otherwise a
+      map with plural forms as keys (`0`, `1` and so on) and plural strings as
+      corresponding values
 
   """
 
   @type t :: %__MODULE__{
     msgid: binary,
-    msgstr: binary,
+    msgid_plural: binary,
```
josevalim: Given those changes, maybe two structs, Translation and PluralTransation, would be better? It would also allow us to skip some nil checks in other places... 

For example, the fact msgstr is a binary or map can be a smell.

josevalim:
This looks really, really, really good!

lib/gettext/pluralizer.ex
```diff
@@ -0,0 +1,40 @@
+defmodule Gettext.Pluralizer do
```
josevalim: Maybe we can define the behaviour in `Gettext.Plural` itself and says the module already ships with an implementation? That would be my favorite approach.

lib/gettext.ex
```diff
@@ -63,13 +63,26 @@ defmodule Gettext do
...
   end
 
+  @doc false
+  def pluralize(locale, msgstr, n) do
+    plural_form = Gettext.Plural.plural(locale, n)
+    translation = Map.get(msgstr, plural_form)
+    {:ok, replace_count(translation, n)}
+  end
+
+  defp replace_count(string, n) do
```
josevalim: We need to generalize this. Let's add support for interpolation as the next feature. You can leave a TODO for now to do the interpolation at compile time and we will tackle it next.

lib/gettext/pluralizer.ex
whatyouhide: That's exactly what I did as I think it's the nicest thing as well, but I kept getting a warning for an undefined behaviour. Something I'm missing?

josevalim: You can have that but you will have to skip `@behaviour Gettext.Plural` in `Gettext.Plural` itself.

whatyouhide:
@josevalim I made the changes you suggested, let me know what you think. Thanks for reviewing btw!

josevalim:
Btw, any idea why the build failed? Maybe we should test only against 1.0.2?

josevalim:
Ship it!  :shipit:

lexmag:
Regarding Travis CI failures, it is a problem with their `language: elixir` support (still beta), it helps to use previous configuration that uses precompiled elixir. :bowtie:

whatyouhide:
Ok, merging this then. I'll revert to the precompiled Elixir for Travis :)

PR #10, opened 2015-04-16
Interpolation

whatyouhide:
Support for interpolation, yay!

This is by far the "dirtiest" code I had to write within gettext, but I had some trouble finding cleaner ways of doing things. One of the dirtiest things I did is using `Code.eval_quoted/2` ([here](https://github.com/whatyouhide/gettext/blob/interpolation/lib/gettext.ex#L134)), but I couldn't come up with a better way to do this. Another smelly thing is that I shared a variable (`bindings`) between macros, effectively making them *not* hygienic.

I'll wait for any (very welcome!) feedback or suggestions :). Thanks in advance!

### DIFF
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -55,7 +55,16 @@ defmodule Gettext do
       @before_compile Gettext
 
       def lgettext(locale, domain, msgid, bindings \\ %{})
+
+      def lgettext(locale, domain, msgid, bindings) when is_list(bindings) do
+        lgettext(locale, domain, msgid, Enum.into(bindings, %{}))
+      end
+
       def lngettext(locale, domain, msgid, msgid_plural, n, bindings \\ %{})
+
+      def lngettext(locale, domain, msgid, msgid_plural, n, bindings) when is_list(bindings) do
+        lngettext(locale, domain, msgid, msgid_plural, n, Enum.into(bindings, %{}))
+      end
     end
   end
 
@@ -78,7 +87,7 @@ defmodule Gettext do
       end
 
       def lngettext(_, _, msgid, msgid_plural, n, bindings) do
-        str = if n == 1, do: msgid, else: msgid_plural
+        str      = if n == 1, do: msgid, else: msgid_plural
         bindings = Map.put(bindings, :count, n)
 
         case Gettext.Interpolation.interpolate(str, bindings) do
@@ -147,18 +156,12 @@ defmodule Gettext do
     interpolation = compile_interpolation(str)
 
     quote do
-      bindings = var!(bindings)
-
-      if is_list(bindings) do
-        bindings = Enum.into bindings, %{}
-      end
-
-      case bindings do
+      case var!(bindings) do
         unquote(match) ->
           {:ok, unquote(interpolation)}
         _ ->
           keys = unquote(keys)
-          {:error, Gettext.Interpolation.missing_interpolation_keys(bindings, keys)}
+          {:error, Gettext.Interpolation.missing_interpolation_keys(var!(bindings), keys)}
       end
     end
   end
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -43,6 +43,7 @@ defmodule Gettext do
 
   alias Gettext.PO.Translation
   alias Gettext.PO.PluralTranslation
+  alias Gettext.Interpolation
 
   @default_priv "priv/gettext"
 
@@ -92,6 +93,8 @@ defmodule Gettext do
     :filelib.fold_files(dir, "\.po$", true, &compile_po_file(&1, &2), [])
   end
 
+  # `acc` is a list of already compiled translation, i.e., of quoted function
+  # definitions.
   defp compile_po_file(path, acc) do
     {locale, domain} = locale_and_domain_from_path(path)
     translations     = Gettext.PO.parse_file!(path)
@@ -108,14 +111,12 @@ defmodule Gettext do
   end
 
   defp compile_translation(locale, domain, %Translation{} = t) do
-    bindings_match = compile_bindings_match(t.msgstr)
+    bindings_match        = compile_bindings_match(t.msgstr)
     interpolated_bindings = compile_interpolated_bindings(t.msgstr)
-    required_keys = bindings_match |> elem(2) |> Dict.keys
+    required_keys         = Interpolation.bindings_in_string(t.msgstr)
 
     quote do
       def lgettext(unquote(locale), unquote(domain), unquote(t.msgid), bindings) do
-        import Gettext.Interpolation, only: [missing_interpolation_keys: 2]
-
         if is_list(bindings) do
           bindings = Enum.into(bindings, %{})
         end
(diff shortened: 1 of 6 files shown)


lib/gettext.ex
```diff
@@ -43,14 +43,19 @@ defmodule Gettext do
 
   alias Gettext.PO.Translation
   alias Gettext.PO.PluralTranslation
+  alias Gettext.Interpolation
 
   @default_priv "priv/gettext"
+  @pluralizer Application.get_env(:gettext, :plural_forms, Gettext.Plural)
```
josevalim: `@plural_forms`?

lib/gettext.ex
```diff
@@ -106,27 +111,75 @@ defmodule Gettext do
...
+  # the quoted code that is returned.
+  defp compile_interpolation_code(str) do
+    keys          = Interpolation.keys(str)
+    match         = compile_interpolation_match(keys)
+    interpolation = compile_interpolation(str)
+
+    quote do
+      bindings = var!(bindings)
+
+      if is_list(bindings) do
```
josevalim: Instead of checking in every function, define an earlier clause for `lgettext` and `lngettext` that simply checks if bindings is a list and converts it to a map.

lib/gettext.ex
```diff
@@ -106,27 +111,75 @@ defmodule Gettext do
...
+  defp compile_translation(locale, domain, %PluralTranslation{} = t) do
+    compiled_plural_forms = for {form, str} <- t.msgstr, into: %{} do
+      {form, compile_interpolation_code(str)}
+    end
+
     quote do
-      def lgettext(unquote(locale), unquote(domain), unquote(msgid)) do
-        {:ok, unquote(msgstr)}
+      def lngettext(unquote(locale), unquote(domain), unquote(t.msgid), unquote(t.msgid_plural), n, bindings) do
+        plural_form = unquote(@pluralizer).plural(unquote(locale), n)
```
josevalim: I think you can get rid of the eval if you compile to this:

```elixir
bindings = Map.put(bindings, :count, n)
case unquote(@pluralizer).plural(unquote(locale), n) do
  0 ->
    case bindings do
      unquote(match) -> mesage_for_0
      _ -> keys
    end
  1 -> ...
  ...
end
```

Inside each pattern in the first case, you will have exactly the same compiled pattern as in the regular interpolation.

lib/gettext.ex
```diff
@@ -106,27 +111,75 @@ defmodule Gettext do
...
+  # string based on some bindings or returns an error in case those bindings are
+  # missing. Note that the `bindings` variable is assumed to be in the scope by
+  # the quoted code that is returned.
+  defp compile_interpolation_code(str) do
+    keys          = Interpolation.keys(str)
+    match         = compile_interpolation_match(keys)
+    interpolation = compile_interpolation(str)
+
+    quote do
+      bindings = var!(bindings)
```
josevalim: Avoid this assignment, just call `var!(bindings)` where you need.

lib/gettext/interpolation.ex
```diff
@@ -0,0 +1,105 @@
...
+  ## Examples
+
+      iex> interpolate "Hello %{name}", name: "José"
+      {:ok, "Hello José"}
+      iex> interpolate "%{count} errors", %{name: "Jane"}
+      {:error, "missing interpolation keys: count"}
+
+  """
+  @spec interpolate(binary, Dict.t) :: {:ok, binary} | {:error, binary}
+  def interpolate(str, bindings) do
```
josevalim: This is function is not efficient because you are calling to_interpolatable twice. One in `keys` and another time down below. Can it be rewritten to maybe traverse the whole string just once?

whatyouhide:
@josevalim I should have covered everything you pointed out with these new commits. Let me know what you think and thanks so much for the feedback!

lib/gettext.ex
whatyouhide: Very neat solution, wow! It worked almost verbatim, I just had to add a `var!(bindings) = Map.put(...)` and it worked :).

lib/gettext/interpolation.ex
```diff
@@ -0,0 +1,113 @@
...
+    segments = to_interpolatable(str)
+    keys     = keys(segments)
+
+    if keys -- Dict.keys(bindings) != [] do
```
josevalim: Use `Map` instead of `Dict` if we are sure it is a map (and we are).

josevalim: There are other places in this module we could do the same!

josevalim:
I have added one extra comment and I think this is ready to go! What is next?

lib/gettext/interpolation.ex
whatyouhide: @josevalim I pushed a commit where I fix this, but actually I only could do this in [one place](https://github.com/elixir-lang/gettext/blob/c2320e4ebe160c82eb5333e071f9995e392e0b4e/lib/gettext.ex#L139). I don't think we should do that in the `interpolate/2` function since that function is dynamic, so it doesn't need pattern matching on a map and can both accept keyword lists as well as maps (actually, any `Dict.t` type). Wdyt?

josevalim: The interpolate function is private to gettext, we don't expect folks to be using it. If we are, it is worth assuming they are giving maps as well. We don't need to make the code generic if we don't need to. :)

whatyouhide:
@josevalim pushed a new commit using `Map` everywhere, your point was very valid (`Interpolation` being internal to gettext) :).

josevalim:
:shipit: I will write something about the next steps, but this is almost ready for prime time I think.

whatyouhide:
@josevalim I'd like the next step to be a very tiny one that I can easily do while you write stuff about the next steps: default-domain functions! Today, we have `lgettext` and `lngettext`, but I'd like to change them to `ldgettext` and `ldngettext` and re-implement `lgettext` as `ldgettext(locale, "default", ...)` and `lngettext` as `ldngettext(local, "default", ...)`. What do you think? As a side note, this is [what Python does](https://docs.python.org/3/library/gettext.html).

josevalim:
@whatyouhide hold on to that thought because that's exactly what I want to discuss next. The functions we have today are not meant to be called directly (I think). So we need to decide the prime time API, then we can talk about how and why changing `lgettext`. Does it make sense?

whatyouhide:
Yes it does. If we want to make those functions "private" to gettext we
could even call those functions like __this__. I'll wait for the next step
then, and then we can go on from that :).

On Friday, April 17, 2015, José Valim <notifications@github.com> wrote:

> @whatyouhide <https://github.com/whatyouhide> hold on to that thought
> because that's exactly what I want to discuss next. The functions we have
> today are not meant to be called directly (I think). So we need to decide
> the prime time API, then we can talk about how and why changing lgettext.
> Does it make sense?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/elixir-lang/gettext/pull/10#issuecomment-93969129>.
>


-- 

Andrea Leopardi
an.leopardi@gmail.com

PR #12, opened 2015-04-19
Refactor `Gettext` by creating `Gettext.Compiler`

whatyouhide:
I moved *all* of the compilation logic (which transformed `.po` files into
function clauses) used in the `Gettext` module to the `Gettext.Compiler` module.
That's it.

I didn't change the tests **at all**, since I don't know if it's wise to test the compilation step instead of just testing the end result (by testing the compiled functions in a module that uses `Gettext`). Let me know what you think on this!

I would also like some feedback on how I performed the refactoring: do you think we should refine what the `Gettext.Compiler` module does (e.g., it only compiles `.po` files to `case` expressions and then `Gettext` actually injects `lgettext/4` and `lngettext/6` in the target module) or we can just delegate everything to `Gettext.Compiler`?

Thanks for the input!


lib/gettext.ex
```diff
@@ -76,110 +61,9 @@ defmodule Gettext do
...
-  # `%{var}` patterns are turned into `var` variables, namespaced inside the
-  # current `__MODULE__`. Heavily inspired by Chris McCord's "linguist", see
-  # https://github.com/chrismccord/linguist/blob/master/lib/linguist/compiler.ex#L70
-  defp compile_interpolation(str) do
-    Enum.reduce Interpolation.to_interpolatable(str), "", fn
-      key, acc when is_atom(key) ->
-        quote do: unquote(acc) <> to_string(unquote(Macro.var(key, __MODULE__)))
-      str, acc ->
-        quote do: unquote(acc) <> unquote(str)
+      unquote(Compiler.function_signatures)
```
josevalim: Here is too late for the functional signatures.

lib/gettext.ex
```diff
@@ -41,30 +41,15 @@ defmodule Gettext do
...
+  alias Gettext.Compiler
 
   @default_priv "priv/gettext"
-  @plural_forms Application.get_env(:gettext, :plural_forms, Gettext.Plural)
 
   @doc false
   defmacro __using__(opts) do
     quote do
       @gettext_opts unquote(opts)
       @before_compile Gettext
```
josevalim: I would honestly just do `@before_compile Gettext.Compiler` and just expose `signatures/0` and `__before_compile__` in Compiler.

whatyouhide:
@josevalim thanks for the feedback, I made the changes you pointed out and `push --force`'d.

If we're ready to merge, there's just one last thing I'd like to discuss: since we're moving towards the direction of using `lgettext/4` and `lngettext/6` only as "private" functions, maybe we could drop the support for bindings as keyword lists in `lgettext/4` and `lngettext/6` and move it up to where we'll call these functions. This way, we can simplify the code a little bit since we're always sure we'll get bindings as maps. What do you think?

josevalim:
> This way, we can simplify the code a little bit since we're always sure we'll get bindings as maps. What do you think?

Up to you. Better to wait until we are sure so we don't have any rework.

whatyouhide:
Ok then, merging this :).

PR #13, opened 2015-04-21
Public API

whatyouhide:
I have a working (albeit a bit rough) version of the public API for Gettext, as per #11.
I want to make sure I got everything right:

* `Gettext.d?n?gettext` has to handle responses from the backend like `{:ok, string}`, `{:default, string}` and `{:error, error}`. If I get an `:ok` or a `:default` I return the string, otherwise I raise, right?
* The `d?n?gettext` macros generated in each backend must only handle `{:ok, string}` responses for now, right? I guess `:error`s should be raised and we should warn on `:default` responses, right?

I'm holding off until I'm sure I'm in the right direction; I still don't implement the `Gettext.d?n?gettext` variants which take a locale as an argument, even if that should be super-quick to do.

Comments? :)

### DIFF
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -87,4 +87,19 @@ defmodule Gettext do
   def gettext(backend, string, bindings \\ %{}) do
     dgettext(backend, "default", string, bindings)
   end
+
+  @spec dngettext(atom, binary, binary, binary, non_neg_integer, Map.t) :: binary
+  def dngettext(backend, domain, id, plural_id, n, bindings \\ %{}) do
+    case backend.lngettext(locale(), domain, id, plural_id, n, bindings) do
+      {atom, string} when atom in [:ok, :default] ->
+        string
+      {:error, error} ->
+        raise Gettext.Interpolation.MissingKeysError, error
+    end
+  end
+
+  @spec ngettext(atom, binary, binary, non_neg_integer, Map.t) :: binary
+  def ngettext(backend, id, plural_id, n, bindings \\ %{}) do
+    dngettext(backend, "default", id, plural_id, n, bindings)
+  end
 end
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -43,8 +43,6 @@ defmodule Gettext do
 
   @type locale :: binary
 
-  @default_locale "en"
-
   @doc false
   defmacro __using__(opts) do
     quote do
@@ -73,7 +71,7 @@ defmodule Gettext do
     if locale = Process.get(__MODULE__) do
       locale
     else
-      default_locale = Application.get_env(:gettext, :default_locale, @default_locale)
+      default_locale = Application.get_env(:gettext, :default_locale)
       Process.put(__MODULE__, default_locale)
       default_locale
     end
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -55,7 +55,15 @@ defmodule Gettext do
   end
 
   @spec locale() :: locale
-  def locale, do: Process.get(__MODULE__, @default_locale)
+  def locale do
+    if locale = Process.get(__MODULE__) do
+      locale
+    else
+      default_locale = Application.get_env(:gettext, :default_locale, @default_locale)
+      Process.put(__MODULE__, default_locale)
+      default_locale
+    end
+  end
 
   @spec locale(locale) :: nil
   def locale(locale) when is_binary(locale),
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -70,4 +70,21 @@ defmodule Gettext do
     do: Process.put(__MODULE__, locale)
   def locale(_locale),
     do: raise(ArgumentError, "locale/1 only accepts binary locales")
+
+  @spec dgettext(atom, binary, binary, Map.t) :: binary
+  def dgettext(backend, domain, string, bindings \\ %{}) do
+    case backend.lgettext(locale(), domain, string, bindings) do
+      {:ok, string} ->
+        string
+      {:default, string} ->
+        string
+      {:error, error} ->
+        raise Gettext.Interpolation.MissingKeysError, error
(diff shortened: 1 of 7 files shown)


lib/gettext.ex
```diff
@@ -49,4 +53,76 @@ defmodule Gettext do
...
+      Gettext.locale()
+      #=> "en"
+
+  """
+  @spec locale() :: locale
+  def locale do
+    if locale = Process.get(__MODULE__) do
+      locale
+    else
+      default_locale = Application.get_env(:gettext, :default_locale, @default_locale)
```
josevalim: You don't need `@default_locale`. Go to the mix.exs and add `env: [default_locale: :en]` inside `def application` and now we are sure something sensible will be set.

lib/gettext.ex
```diff
@@ -49,4 +53,76 @@ defmodule Gettext do
...
+  end
+
+  @spec ngettext(atom, binary, binary, non_neg_integer, Map.t) :: binary
+  def ngettext(backend, id, plural_id, n, bindings \\ %{}) do
+    dngettext(backend, "default", id, plural_id, n, bindings)
+  end
+
+  defp handle_backend_result({atom, string}) when atom in [:ok, :default],
+    do: string
+  defp handle_backend_result({:error, error}),
```
josevalim: Theoretically, we don't know the error is about missing interpolation keys. So I would rather raise `Gettext.Error` or make the interpolation key a special one `{:missing_interpolation_keys, keys}`. If we are doing the latter, we still need to support the generic error (imo).

lib/gettext/compiler.ex
```diff
@@ -16,12 +16,69 @@ defmodule Gettext.Compiler do
...
     end
   end
 
+  defp macros do
+    quote unquote: false do
+      defmacro dgettext(domain, msgid, bindings \\ Macro.escape(%{}))
+
+      defmacro dgettext(domain, msgid, bindings) when is_binary(msgid) do
+        quote do
+          locale = Gettext.locale
```
josevalim: No need for the variable. Just call `Gettext.locale` directly. Simpler, faster, cleaner. :)

lib/gettext/compiler.ex
```diff
@@ -16,12 +16,69 @@ defmodule Gettext.Compiler do
...
+      unquote(macros)
       unquote(binding_conversion_clauses)
       unquote(compile_po_files(translations_dir))
       unquote(dynamic_clauses)
     end
   end
 
+  defp macros do
+    quote unquote: false do
+      defmacro dgettext(domain, msgid, bindings \\ Macro.escape(%{}))
```
josevalim: `binding \\ %{}` would be fine.

lib/gettext/compiler.ex
```diff
@@ -16,12 +16,69 @@ defmodule Gettext.Compiler do
...
+
+      defmacro dgettext(domain, msgid, bindings) when is_binary(msgid) do
+        quote do
+          locale = Gettext.locale
+          unquote(__MODULE__).lgettext(locale, unquote(domain), unquote(msgid), unquote(bindings))
```
josevalim: Ah, those need to return strings. So I would rather have it call: `Gettext.dgettext(unquote(__MODULE__), ...)` and so on. What do you think? We could make all functions here just call their `Gettext` counterparts, except they have guard checks.

josevalim: Another idea: if we make everything go through the `Gettext` module, we can also make the conversion of the binding from a list to map inside `Gettext` instead of adding them to `lgettext`. Thoughts? Then `lgettext` does not need optional arguments nor the conversion because it is always called from `Gettext`.

whatyouhide: I know the implementations of the `Gettext.d?n?gettext` functions  and `MyApp.Gettext.d?n?gettext` macros look very similar *now*, but I didn't implement what you suggested for a single reason: I think the `d?n?gettext

lib/gettext.ex
whatyouhide: Agreed, maybe a generic `Gettext.Error` is better for now :).

lib/gettext/compiler.ex
whatyouhide: I think that doesn't work? This is the minimal case I can reproduce:

```elixir
defmodule M do
  defmacro m(arg \\ %{}) do
    quote do
      unquote(arg)
    end
  end
end

require M
M.m(%{}) #=> %{}
M.m      #=> invalid quoted expression: %{}
```

lib/gettext.ex
whatyouhide: Nice!

whatyouhide:
Hey @josevalim, I just fixed the `Gettext.locale()` thing you mentioned, the `:default_locale` in the env and introduced `Gettext.Error`. I responded to the diff comments for the other things :).

whatyouhide:
7192912 fixes the bug found by @josevalim where this didn't work:

```elixir
MyApp.Gettext.gettext @module_attr
```

Using `Macro.expand/2` makes it work.

lib/gettext/compiler.ex
josevalim: You are right. My bad. :)

lib/gettext/compiler.ex
josevalim: We will always ignore the default in `{:default, string}` afaik. At least that's how the regular gettext works and I don't think we should change.

josevalim: We just need to decide this and we are already for merge!

whatyouhide:
Hey @josevalim, fixed the last thing that was missing so I think we're ready to merge this. While we plan the next step, I'll start writing some docs for all the public API introduced with this PR.

I'll wait for your :shipit:

josevalim:
:shipit:

whatyouhide:
:fireworks:

Issue #16, opened 2015-05-04
Comments tokenizing + parsing


whatyouhide:
Comments in Gettext are not like comments everywhere else, since you usually don't just ignore them.

Comments start with an hash (`#`) character and go on until EOL. There are two main kinds of comments:

- translator comments (denoted by `#` followed by some whitespace, most commonly `#[space]`)
- autogenerated comments (`#` followed by some specifier character)

We will need to parse, tokenize and keep **all** those comments associated with each translation since they are necessary to rebuild PO files exactly like we found them. Also, we can obviously make use of the autogenerated comments.

## Translator comments

```po
# This is a translator comment.
msgid "foo"
msgstr "bar"
```

We'll keep them as is, nice and easy.

## Autogenerate comments

```po
#. extracted-comments
#: reference
#, flag
#| msgid previous-untranslated-string
```

**Extracted comments**: they're written by the programmer, in the source code, for the translator. I think we should avoid implementing this for now as we would need some kind of macro to do this as well.

```po
#. Hey translator, thanks!
...
```

**Reference**: file or list of files the next translation comes from. This is the easiest kind of comment to implement. If two translations with the same msgid are found, they're merged into one translation and both source files are listed here.

```po
#: my_app/lib/foo.ex:84 my_app/lib/bar.ex:40
...
```

**Flag**: a list of flags/tags/format directives. Mostly used to denote the source language. We'll most likely automatically add `elixir-format` here.

```po
#, elixir-format
...
```

**Fuzzy**: related to the `fuzzy` flag. Let's just leave this alone for now :|

We're going to add a `:comments` field to the `PluralTranslation` and `Translation` structs, I'll think about the data structure for this. I'd go for a simple map or a `Comments` struct, but we have to decide if the original order of comments found in PO files matters to us (when re-creating the PO files back). If it does, we can go with a list of tagged tuples like `[{:reference, ...}, {:flags, ...}]`. We'll see :)

josevalim:
This sounds great. Do we need to use flag? I think we can safely skip it... otherwise it is just noise in the .po files.

whatyouhide:
I'll see how po editors behave when there's no flag comment. We could actually also use it to easily detect translations we generated instead of having to read the reference comment and search for `.ex` files.

whatyouhide:
[...]

PR #18, opened 2015-05-09
Comments support

whatyouhide:
Hey there, I quickly added support for tokenizing and parsing comments as per #16. I'm opening the PR so that you're all notified but I'm merging it right away since the code is very simple. If you have any comments or feedback I still would appreciate it a lot, and we can change the code later if necessary :). Thanks!

### DIFF
--- a/lib/gettext/po/plural_translation.ex
+++ b/lib/gettext/po/plural_translation.ex
@@ -26,5 +26,5 @@ defmodule Gettext.PO.PluralTranslation do
     msgstr: %{non_neg_integer => binary},
   }
 
-  defstruct [:msgid, :msgid_plural, :msgstr]
+  defstruct msgid: nil, msgid_plural: nil, msgstr: nil, comments: []
 end
--- a/lib/gettext/po/tokenizer.ex
+++ b/lib/gettext/po/tokenizer.ex
@@ -11,7 +11,10 @@ defmodule Gettext.PO.Tokenizer do
     {:plural_form, line, non_neg_integer} |
     {:msgid, line} |
     {:msgid_plural, line} |
-    {:msgstr, line}
+    {:msgstr, line} |
+    {:comm_translator, line, binary} |
+    {:comm_reference, line, binary}
+
 
   # In this list of keywords *the order matters* because a function clause is
   # generated for each keyword, and keywords have to be followed by whitespace.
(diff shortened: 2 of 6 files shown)


josevalim:
This looks great! I have a suggestion for you to keep in mind. You don't need to change anything but it may be handy in the future.

One option is to keep the `:comments` entry as a list of binaries and store the reference directly in the translation? Why? Because it will make easier to match on the reference later and it will also make it easier to dump the comments back.

So instead of having:

    comments: [{:translator, "Hello"}, {:reference: "lib/foo.ex:23"}]

I would have:

    comments: ["# Hello", "#: lib/foo.ex"], reference: {"lib/foo.ex", 23}

I guess we could have multiple references too... so it may need to be:

    comments: [...], references: [{"lib/foo.ex", 23}]

If that is the case, then we don't have large benefits compared to the current approach. In any case, it is worth discussing. :)

whatyouhide:
Thanks for the feedback! So, the first idea of storing `:reference` in the translation could be very nice:

> `comments: ["# Hello", "#: lib/foo.ex"], reference: {"lib/foo.ex", 23}`

The main problem I see with it is that we're pretty much *sure* that there will be some translations with more than on reference, which brings us to the second suggestion:

> `comments: [...], references: [{"lib/foo.ex", 23}]`

This is nice (maybe nice than the solution I implemented), but I can see one main downside: what if we add flag comments or other comment types? This solution isn't "scalable" in that sense, right?

Your solution has the advantage of being easier to dump back (which is not very hard to do even with a `Keyword.t` list of comments) and of being pattern-matchable.

For now I went with a simple kw list because, well, it's simple :stuck_out_tongue: and it allows us to keep the order of comments but at the same time retrieve all comments of a given type (`Keyword.get_values/2`).

So, what do you think?

josevalim:
Right, the biggest downside of the current approach is that dumping needs to know about the existing comment types. So, in this sense, both aren't "scalable" because once we add a new comment type, like flag, we would need to teach the dumper how to dump it too.

whatyouhide:
Yes, that is true but we wouldn't need to change the `(Plural)Translation` structs in to add a `:flags` field and whatnot. Not modifying the structs in this way would make, e.g., adding a flag very uncomfortable I think. I keep mentioning flags comments because even if we're not implementing them for now, we could need them in order to mark translations as "fuzzy" at some point, which would mean adding a *fuzzy* comment type and you see where this is going :).

josevalim:
Yes, I get it. :) Alternatively, someone would just extract the flag from the comments too, it should not be hard:

    flags = for <<"#&", rest::binary>> <- translation.comments, do: rest

It is up to you!

whatyouhide:
We could provide functions like that ourselves if we need to, but I think users would rarely need to use them. Anyways, let's go with your solution for now, we can always change it later if we'll need to :)

PR #19, opened 2015-05-10
`%Gettext.PO{}` struct

whatyouhide:
Hey there @josevalim, I implemented the `%Gettext.PO{}` struct we talked about some time ago, can you take a quick look at this to see if you had something like it in mind?

What I tried to do is separating responsibilities as much as possible: I think `Gettext.PO.Parser` shouldn't know about `%Gettext.PO{}`s, so I made `Gettext.PO.Parser.parse/1` return a tuple like `{:ok, headers, translations}` instead of `{:ok, translations}`. An instance of `%Gettext.PO{}` is built up in `Gettext.PO.parse_(string|file)` so that only `PO` knows about `%PO{}`. Wdyt?

### DIFF
--- a/lib/gettext/compiler.ex
+++ b/lib/gettext/compiler.ex
@@ -1,6 +1,7 @@
 defmodule Gettext.Compiler do
   @moduledoc false
 
+  alias Gettext.PO
   alias Gettext.PO.Translation
   alias Gettext.PO.PluralTranslation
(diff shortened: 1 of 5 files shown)


josevalim:
Beautiful.  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #20, opened 2015-05-11
Dumping of `%Gettext.PO{}` structs

whatyouhide:
Hey there, I have a first working version of `Gettext.PO.dump/1` which dumps `%Gettext.PO{}` structs to strings. The implementation is straightforward. Don't pay attention to the lack of comments, specs, or docs, I just wanted some feedback on the code before adding those things and merging the PR.

So, we know most translators don't edit PO files manually but instead use PO editors (like [POEdit](http://poedit.net) or [poeditor.com](https://poeditor.com)), so instead of coming up with PO files my self to test the `dump(parse(po)) == po` process, I just generated files with some of those editors and tested the process on them. I only did this with a couple of PO editors, but we can add more in the future. What do you think of this approach? Is it useless since we're always going to modify those files a little bit? I wanted the parse+dump process to change the PO files as little as possible so people don't have to constantly check automatically generated changes into VCS.


whatyouhide:
@josevalim btw, do you think we should use the more structured EEx in order to dump translation or we can leave it simple and just build strings like I did in this PR? EEx may give us a big advantage for its clarity, but also a bit of a mental overhead I think.

test/fixtures/po_editors/poedit.po
```diff
@@ -0,0 +1,25 @@
+msgid ""
```
josevalim: What a horrible hack are the PO headers :P

lib/gettext/po.ex
```diff
@@ -128,4 +130,54 @@ defmodule Gettext.PO do
         raise SyntaxError, file: path, line: line, reason: reason
     end
   end
+
+  def dump(%PO{headers: [], translations: translations}) do
```
josevalim: So, if you make dump return iolists, it is likely the code will be cleaner and faster.

josevalim: One easy way to go about this is to make each entry in a list be a line of the po file.

test/fixtures/po_editors/poedit.po
whatyouhide: Yeah, I know :expressionless: makes the code a little bit uglier as well, but we pretty much can't do anything about it :(

whatyouhide:
Hey @josevalim, I tried to implement your suggestion about iolists but I don't think I've had much success. I'm not extremely confident with using iolists (I got how they work, I just don't know how to put them to good use :smile:) and I think the code may reflect that. You suggested to make each entry in the final iolist a line in the file, but doesn't that go against the point of using iolists? To make each entry in the list a line in the file I had to abuse the `++` operator, since we inherently have nested data structures (e.g., a list of translations, each of which has a list of headers and so on). So I went with nesting iolists as well, which comes very natural at this point; however, I'm not sure the code is more readable. It may be faster, not that we care *that* much since this stuff is meant to happen at compile time. Wdyt?

lib/gettext/po.ex
```diff
@@ -128,4 +130,67 @@ defmodule Gettext.PO do
...
+    msgstr ""
+    """
+
+    [base|Enum.map(headers, &(~s("#{&1}\\n"\n)))]
+  end
+
+  defp dump_translations(translations) do
+    translations
+    |> Enum.map(&dump_translation/1)
+    |> Enum.intersperse(?\n)
```
josevalim: Remove intersperse here by adding `?\n` downstream. :)

lib/gettext/po.ex
```diff
@@ -128,4 +130,67 @@ defmodule Gettext.PO do
         raise SyntaxError, file: path, line: line, reason: reason
     end
   end
+
+  def dump(%PO{} = po) do
+    po
+    |> do_dump
+    |> IO.iodata_to_binary
```
josevalim: You don't need to convert to iodata to binary here. We want to write to files and files accept io lists. You convert them to binaries in tests...

lib/gettext/po.ex
```diff
@@ -128,4 +130,67 @@ defmodule Gettext.PO do
...
+
+  defp dump_translations(translations) do
+    translations
+    |> Enum.map(&dump_translation/1)
+    |> Enum.intersperse(?\n)
+  end
+
+  defp dump_translation(%Translation{} = t) do
+    translation = """
+    msgid "#{t.msgid}"
```
josevalim: Should we escape quotes inside `t.msgid`?

lib/gettext/po.ex
```diff
@@ -128,4 +130,67 @@ defmodule Gettext.PO do
...
+    ids = """
+    msgid "#{t.msgid}"
+    msgid_plural "#{t.msgid_plural}"
+    """
+
+    [dump_comments(t.comments), ids, dump_plural_msgstr(t.msgstr)]
+  end
+
+  defp dump_comments(comments) do
+    Enum.map comments, &[&1, ?\n]
```
josevalim: It could be: `Enum.map comments, &[&1|?\n]`

josevalim:
Hey, to me the implementations looks definitely better here! I have added some tiny comments! Great work!

whatyouhide:
@josevalim thanks for the feedback! I'm going to work on these in a few hours and merge afterwards so that we can finally move to the automatic generation of PO(T) files!

lib/gettext/po.ex
whatyouhide: That would be an `ArgumentError` since `?\n` is not a list. We could do `&[&1|'\n']`, would it be faster than creating a list with two elements?

josevalim: It isn't an argument error. :) That is called an improper list (when a list does not terminate with an empty list). It takes less memory because we use a cons cell instead of a list. Basically, instead of:

    [comment|[?\n|[]]]

We have just:

    [comment|?\n]

whatyouhide: Wow, well I look like an idiot now :). I didn't try that in IEx, just got an argument error and assumed it was that. Sorry :disappointed: At least I learned something new :).

lib/gettext/po.ex
whatyouhide: I used `Enum.intersperse/2` instead of appending `?\n` to each translation since that way the last translation would have an additional newline (the PO file would end with `\n\n`). Wdyt?

josevalim: Ah, I forgot about this detail. :shipit:

lib/gettext/po.ex
whatyouhide: Good catch! In the new commits I pushed I'm also escaping `\n`, `\r` and `\t`. :)

whatyouhide:
Merging this, on to the merging of PO(T) files now!

Issue #17, opened 2015-05-04
Generating and merging PO(T) files


whatyouhide:
Opening this issue to lay down what @josevalim and I discussed yesterday. So, the next part in the development of Gettext for Elixir will be focused on compile-time niceties. These include automatic generation of POT and PO files as well as merging of such files.

First of all, just as a reference, POT files are just like PO files but with empty translations. Example of snippet of POT file:

```po
msgid "Hello there!"
msgstr ""

msgid "One error"
msgid_plural "%{count} errors"
msgstr[0]
msgstr[1]
```

The first step for implementing all of this will be to create a `Gettext.PO` struct that will represent a PO file. We need this since PO files contain translations as well as metadata (headers at the top of the file).

After we have a `Gettext.PO` struct in place, we'll create the code that dumps an instance of this struct as a PO(T) file. The goal (and the ultimate test) will be to make this true:

```elixir
dump(parse(file)) == file
```

## Merging PO(T) files

Once this is done, the next step will be to write code that *merges* two PO(T) files. This is probably the trickiest thing to do. PO(T) files will be merged every time the user runs whatever task we'll provide and some PO(T) files already exist. During this step, we'll do this:

* add all the new translations found in the source code
* remove all the old translations that are not in the source code anymore; we'll only remove the translations with a `#: reference` comment (see #16 for some info on comments) assuming that we automatically generated them
* leave translations without a reference comment untouched, even if they're not in the source code anymore

## Generating PO files

This will be pretty easy at this point. We just need to tweak the `*gettext` macros we have now so that we know all the translations at compile time.

## The future...

We can probably refine the process of merging PO(T) files, for example by not removing translations based on string similarities.

---

This is everything I remember from our conversation, @josevalim if I missed something add it here so we have everything in one place :).

josevalim:
That sounds great. It is worth mentioning that merging pot and po files are very similar except that:

1. If a translation has no reference in .pot files, we won't remove them

2. If a translation has no reference in .po files, we still remove them, because we assume .pot files is the canonical source

Besides that, the process should be exactly the same.

whatyouhide:
I'm closing this: the only thing left to implement from this issue is the generation of the PO(T) files and I opened #26 to give space to a more focused discussion on that.

PR #21, opened 2015-05-15
Merging PO(T) files

whatyouhide:
So this is the *first* and **very basic** version of merging PO(T) files. I'd like you @josevalim to take a look at this before continuing to work on it since I think your vision will be very useful here :).

Things I'd like to point your attention to:

* I kind of *had* to create a `Gettext.PO.Translations` module in order to keep all the translation-related functions in the same place (we normally would have put such functions in the `Gettext.PO.Translation` module but we have `PluralTranslation` as well :|). The name may be confusing and I'm still not sure about it, but without such a module those kind of functions would have ended up in `Gettext.PO` (as private functions, so no tests for them). This way they're nicely grouped in the same module.
* There are some grey areas when merging PO files, e.g., what do we do with headers? Do we want to keep the translations as much ordered as possible (I think so, VCS would be hell otherwise)?

Note that the code in this PR doesn't make a distinction between merging PO files, POT files or both. We'll want to do that but maybe we should call the process of merging a new POT file with old PO files "updating" instead of merging (e.g., POEdit calls it that way).

Looking forward to feedback :heart:


lib/gettext/po.ex
```diff
@@ -234,4 +235,46 @@ defmodule Gettext.PO do
...
+    translations =
+      discard_autogenerated(unique_old) ++
+      merge_translations(common_old, common_new) ++
+      unique_new
+
+    %PO{translations: translations, headers: merge_headers(old.headers, new.headers)}
+  end
+
+  defp merge_translations(old, new) do
+    for {old_t, new_t} <- Enum.zip(Enum.sort(old), Enum.sort(new)) do
```
josevalim: I am afraid this Enum.sort/1 is going to be accidental and could change in the future when we change the struct. Is there a field we could choose for the sorting explicitly?

lib/gettext/po.ex
```diff
@@ -234,4 +235,46 @@ defmodule Gettext.PO do
...
+  def merge(old, new) do
+    {common_old, unique_old} = Enum.partition old.translations, fn(t) ->
+      Enum.find new.translations, &Translations.same?(t, &1)
+    end
+
+    {common_new, unique_new} = Enum.partition new.translations, fn(t) ->
+      Enum.find old.translations, &Translations.same?(t, &1)
+    end
+
+    translations =
```
josevalim: You mentioned you want to keep the order as much as possible and this commit will move something a bit. Maybe the cleanest way to keep the order is to skip the first partition and use a flat map:

    Enum.flat_map old.translations, fn t ->
      cond do
        same = Enum.find new.translations, &Translations.same?(t, &1) ->
          [merge_translation(t, same)]
        autogenerated?(t) ->
          []
        true ->
          [t]
      end
    end
    
This way we also get rid of the sorting issues i mentioned below.

josevalim:
This looks great! I have added some comments regarding the sorting but everything else looks perfect. :heart:

lib/gettext/po.ex
whatyouhide: Neat!

whatyouhide:
So, I'll work on this tonight but I wanted to write down some thoughts first. This PR is a little messy for what concerns the difference between PO vs POT files. `Gettext.PO.merge/2` as implemented now simply doesn't care about its arguments being PO or POT files. However, as we discussed in #17, merging POT files is slightly different than merging a POT file into a PO file.

I think we should tackle this; we could create a function to specifically merge POT files (which is very similar to the current `merge/2`, except it ignores `msgstr`s since they have to be empty in a POT file), say `merge_pot/2`. After that, we can create a function `merge_pot_into_po/2` which behaves slightly different, i.e., it removes all obsolete translations (the POT is the canonical reference). These two functions could share a bunch of the helper functions that `merge/2` currently uses, so there shouldn't be any plain repetition.

As an alternative to this, we can add a `:template?` (or `:pot?`) field to the `Gettext.PO` struct; the value of this field is determined when the struct is created, either by explicitly setting it or by checking that all the `msgstr`s are empty. This way we can keep only one `merge/2` function and make it polymorphic (since we can tell PO from POT files).

Wdyt @josevalim?

(PS we should give a thought to merging headers too, but that's a story for another day)

lib/gettext/po.ex
whatyouhide: @josevalim implemented your suggestion almost verbatim, works like a charm!

josevalim:
:shipit:

josevalim:
> I think we should tackle this; we could create a function to specifically merge POT files (which is very similar to the current merge/2, except it ignores msgstrs since they have to be empty in a POT file), say merge_pot/2.

I don't think we need this. The msgstrs are going to be empty anyway... which means we don't need to do anything! :)

Btw, we should probably look at gettext for the operation names. Check this out: http://www.gnu.org/software/gettext/manual/gettext.html#index-msgmerge. Our code is almost the same except our code keeps the non autogenerated and theirs do not. So I think the only thing we need to do is to add an option like `:keep_non_autogenerated` (or a better name) and call it a day. We probably don't need separate functions even.

Also, we need to discuss what we are going to do with duplicated entries in the same .po file. What if we have the same msgid twice? It will probably emit a warning when compiling but the warnings won't be clear. We can probably keep a HashDict when we load the translations storing all `msgid => line`, if there is a duplicate, we could store it in a list which we return as result of parse, something like: `{:dup_msgid, [line1, lin2, msgid]}`. Then the entity calling parse can decide if it is going to raise, print errors, etc. This is the approach used by Erlang's compiler.

whatyouhide:
d20947e adds support for the `:keep_manual_translations` option in `Gettext.PO.merge/2`; this option defaults to `true` (current behaviour) but can be set to `false` if we want to discard all obsolete translations (that are not in the reference - PO or POT - file).

This option will likely let us avoid making separate functions for PO/POT files as we can reproduce the behaviour of a newer POT file being merged into an older PO file by setting `:keep_manual_translations` to `false`, meaning the POT file becomes the absolute reference for the merge.

lib/gettext/po.ex
```diff
@@ -234,4 +235,57 @@ defmodule Gettext.PO do
...
   defp escape_char(?\r),  do: ~S(\r)
   defp escape_char(char), do: <<char>>
+
+  @doc """
+  Merges two `Gettext.PO` structs, giving precedence to the second one.
+
+  This function is used to merge two `Gettext.PO` structs into one. `new` has
+  precedence over `old`, meaning that:
+
+    * all headers from `old` are discarded (only the ones from `new` are kept)
```
josevalim: I think we want to keep po headers when merging pot into po and also all comments. I think we already merge comments, right?

whatyouhide: Absolutely. This is how I dealt with headers when I wrote this, quoted directly from the PR 

> (PS we should give a thought to merging headers too, but that's a story for another day)

:) Comments are correctly merged: no translator `#[space]` comment is discarded (they're merged and put one after the other), reference comments from the old file are ditched.

I'm playing with `msgmerge` now and indeed it looks like only headers from PO files are kept (headers in POT files are ignored). Nice and easy :)

whatyouhide:
I fixed the handling of headers in `Gettext.PO.merge/2`, I'm going to finally merge this :).

PR #23, opened 2015-05-24
String wrapping

whatyouhide:
I implemented what @josevalim and I discussed yesterday on IRC: now strings in PO files are not concatenated when parsed, but only when BEAM code is generated. This allows us to dump with no modifications things like this PO file:

```po
msgid "foo"
  "bar"
  "baz"
msgstr "oof"
  "rab"
  "zab"
```

I opened the PR just to notify you of the changes, I'm gonna merge it right away so that I can move on in the meantime :).


whatyouhide:
Btw, I made a choice I'm not 100% sure of: the `:msgid(_plural)?` field in the `%Translation{}` and `%PluralTranslation{}` structs can both be a single string as well as a list of strings: the former happens when there's only a string in the PO file, the latter when there are more than one. This is to avoid the super annoying one elements lists everywhere (`%Translation{msgid: ["foo"], msgstr: ["bar"]}`). I can revert this part if you think it's not simple/elegant, but I figured that since we use `IO.iodata_to_binary/1` on these strings we should be ok. :)

josevalim:
I think it is cleaner to have them always lists. Otherwise folks will assume they are strings and their code will later break. :) Everyhing else is great though!

whatyouhide:
Disclaimer: I didn't want to wrap every string in the test/ directory inside brackets. There's a lot of them! :)

I'll fix this later:)

Andrea

whatyouhide:
@josevalim fixed it, it's on master!

PR #24, opened 2015-05-25
`:po_source` field for translations

whatyouhide:
Opening the PR just to notify you, I'm merging it right away :).

I added the `:po_source` field to both the `Translation` and `PluralTranslation` structs, as discussed with @josevalim on IRC. This field is a `{file, line}` tuple which points to where a translation comes from. For now, the `line` in the tuple will be used by the parser in order to throw meaningful errors when duplicate translations are found; the `file` is still unused, but we may use it later on.

There's actually one thing that bothers me: in order to have a file in the `{file, line}` tuple, we have to loop through all the translations in `Gettext.PO.parse_file/1` because we actually don't have a filename before that step; I don't love this but it's the only *very simple* way I could think of (an alternative would have been to pass the filename down to the parser, but I hate that road :(). All of this, and we don't even **use** the file in `:po_source` :) If any of you has a better idea let me know!

### DIFF
--- a/lib/gettext/po.ex
+++ b/lib/gettext/po.ex
@@ -107,8 +107,13 @@ defmodule Gettext.PO do
   @spec parse_file(Path.t) :: {:ok, t} | parse_error | {:error, atom}
   def parse_file(path) do
     case File.read(path) do
-      {:ok, contents}           -> parse_string(contents)
-      {:error, _reason} = error -> error
+      {:ok, contents} ->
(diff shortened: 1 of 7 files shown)


josevalim:
Another idea is to store the filename in the Gettext.PO and leave it out of the translations. What do you think?

whatyouhide:
@josevalim infinitely better :) Will implement that tomorrow!

PR #25, opened 2015-05-26
Duplicate translations

whatyouhide:
This PR handles duplicate translations. The implementation of the function that checks for duplicates uses `throw`/`catch` in order to traverse the list of translations only one but the code is very simple and straightforward, so that should not be a problem. Thanks @josevalim for the help!

### DIFF
--- a/lib/gettext/po/parser.ex
+++ b/lib/gettext/po/parser.ex
@@ -64,33 +64,27 @@ defmodule Gettext.PO.Parser do
     do: {[], translations}
 
   defp check_for_duplicates(translations) do
-    if duplicates = (translations |> group_translations_by_id |> find_duplicates) do
-      build_duplicate_error(duplicates)
-    else
+    try do
+      Enum.reduce translations, HashDict.new, fn(t, acc) ->
+        id = translation_id(t)
+        {_, line} = t.po_source
+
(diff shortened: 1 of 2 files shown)


lib/gettext/po/parser.ex
```diff
@@ -56,4 +62,29 @@ defmodule Gettext.PO.Parser do
...
+          throw({old_line, line})
+        else
+          Dict.put_new(acc, id, line)
+        end
+      end
+
+      :ok
+    catch
+      {old_line, line} ->
+        {:error, line, "found duplicate of this translation on line #{old_line}"}
```
josevalim: Should we include the msgid in the error message?

josevalim:
:+1

lib/gettext/po/parser.ex
whatyouhide: We could, of course :) do you think very long msgids could be a problem?

josevalim: @whatyouhide yeah, we could show just the first line and truncate with ... if more than one. Another idea is to do this:

    "found duplicate on line #{old_line} for msgid: #{the whole message id}"

If it is at the end, it is not such a big problem to be long.

PR #27, opened 2015-06-01
Extracting translations

whatyouhide:
@josevalim this should be a pretty complete implementation of translation extraction. I left a few comments throughout the commits to discuss a couple of things :)

(docs for the `mix gettext.extract` task are still missing)

### DIFF
--- a/lib/gettext/extractor.ex
+++ b/lib/gettext/extractor.ex
@@ -41,9 +41,13 @@ defmodule Gettext.Extractor do
   """
   @spec process_results() :: :ok
   def process_results do
-    for {backend, domains} <- ExtractorAgent.get_all,
+    # Let's remember to call Map.values/1 since `translations` is a map of
+    # `translation_id => translation`.
+
+    for {backend, domains}     <- ExtractorAgent.get_all,
         {domain, translations} <- domains do
-      merge_or_create_pot_file(backend, domain, translations)
+
+      merge_or_create_pot_file(backend, domain, Map.values(translations))
     end
 
     :ok
@@ -76,21 +80,21 @@ defmodule Gettext.Extractor do
   end
 
   defp pot_file(translations) do
-    # First, let's sort all the translations.
-    translations = Enum.sort_by translations, fn
-      %Translation{msgid: id, references: refs} ->
-        {id, refs}
-      %PluralTranslation{msgid: id, msgid_plural: idp, references: refs} ->
-        {id, idp, refs}
-    end
-
-    # Let's sort the references as well.
-    translations = Enum.map(translations, &(%{&1 | references: Enum.sort(&1.references)}))
+    # Sort all the translations and the references of each translation in order
+    # to make as few changes as possible to the PO(T) files.
+    translations =
+      translations
+      |> Enum.sort_by(&Gettext.PO.Translations.key/1)
+      |> Enum.map(&sort_references/1)
 
     %Gettext.PO{translations: translations}
   end
 
   defp pot_path(backend, domain) do
     Path.join(backend.__gettext_dir__(), "#{domain}.pot")
+  end
+
+  defp sort_references(translation) do
+    update_in(translation.references, &Enum.sort/1)
   end
 end
--- a/lib/gettext/extractor.ex
+++ b/lib/gettext/extractor.ex
@@ -1,0 +1,96 @@
+defmodule Gettext.Extractor do
+  @doc false
+
+  alias Gettext.ExtractorAgent
+  alias Gettext.PO.Translation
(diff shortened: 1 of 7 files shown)


lib/gettext/extractor_agent.ex
```diff
@@ -0,0 +1,39 @@
...
+      end
+    end
+  end
+
+  def get_all do
+    Agent.get __MODULE__, &(&1)
+  end
+
+  defp add_translation_with_merge(translations, translation) do
+    existing = Enum.find(translations, &Gettext.PO.Translations.same?(&1, translation))
```
josevalim: This is going to be expensive because we are going to traverse the whole thing for every new translation, converting the ids to binary. My proposal is to provide something like: `Gettext.Po.Translations.key` that generates a dict key for the translation, using the msgid and the msgid_plural. This way the translations can be stored in a dictionary too and this will be much faster. We lose ordering but I think tit is fine as we order later anyway.

lib/gettext/extractor.ex
```diff
@@ -0,0 +1,96 @@
...
+      new_pot = Gettext.PO.merge(old_pot, new_pot)
+    end
+
+    File.write!(pot_path, Gettext.PO.dump(new_pot))
+  end
+
+  defp pot_file(translations) do
+    # First, let's sort all the translations.
+    translations = Enum.sort_by translations, fn
+      %Translation{msgid: id, references: refs} ->
```
josevalim: Can we have duplicated ids? I don't think we can... so the refs here are useless?

lib/gettext/extractor.ex
```diff
@@ -0,0 +1,96 @@
+defmodule Gettext.Extractor do
```
josevalim: Minor nitpicking: I would call it Collector because we are not really extracting... we are collecting them while the code compiles. Wdyt?

lib/gettext/extractor_agent.ex
```diff
@@ -0,0 +1,39 @@
+defmodule Gettext.ExtractorAgent do
```
josevalim: Ideally we should run this inside a supervision tree. The supervisor in gettext by default will have no children, unless the collecting option is set.

josevalim: The current setup is ok for testing though. We can start the extractor manually in tests.

lib/mix/tasks/gettext.extract.ex
```diff
@@ -0,0 +1,19 @@
...
+  def run(_args) do
+    Mix.shell.info "About to extract Gettext translation from source. Recompiling..."
+
+    Gettext.Extractor.setup_for_extraction
+    force_compile
+    Gettext.Extractor.process_results
+  end
+
+  defp force_compile do
+    Enum.each ~w(compile compile.all compile.elixir), &Mix.Task.reenable/1
```
josevalim: This shouldn't be needed.

josevalim: Ah, I see why you called things Extractor... because it makes more sense from the user perspective. Maybe we can rename this task? `gettext.generate_pot`? How does gettext usually call this?

whatyouhide: GNU gettext uses the `xgettext` where I guess `x` stands for "extract":

> The xgettext program extracts translatable strings from given input files.

I named everything collect-* first, then I renamed everything to extract-* because of this :P I can change it back without problems.

lib/gettext/extractor.ex
whatyouhide: Yep, good catch :) thanks!

lib/mix/tasks/gettext.extract.ex
josevalim: Let's go with extract then!

lib/gettext/extractor_agent.ex
whatyouhide: Much better solution. Will implement soon :)

whatyouhide:
@josevalim made the changes discussed in the comments above, you can review them when you have some time. Btw, removing the `Mix.Task.reenable/1` call on `compile`, `compile.app`, and `compile.elixir` (actually on any one of them) doesn't work for me: nothing is recompiled when I programmatically call `Mix.Task.run("compile", ["--force"])` :(.

whatyouhide:
Pushed some basic tests to  e76079b. The testing strategy is discussed in that commit's message; basically we compile a module that calls `*gettext` functions with `Code.compile_string/2` so that translations are extracted and we can look at them. I didn't test the actual writing of POT files as I still have to think of a clean way to do that.

PR #30, opened 2015-09-08
Add Gettext.with_locale/2

### LINKED ISSUE #28: Add Gettext.with_locale

whatyouhide:
This function runs a given function with the process locale set to the given locale. Useful in many situations, tests included.

Closes #28.

### DIFF
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -482,6 +482,41 @@ defmodule Gettext do
     dngettext(backend, "default", msgid, msgid_plural, n, bindings)
   end
 
+  @doc """
+  Runs `fun` with the gettext locale set to `locale`.
+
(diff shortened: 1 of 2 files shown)


lib/gettext.ex
```diff
@@ -482,6 +482,38 @@ defmodule Gettext do
...
+      end
+      #=> "Ciao mondo"
+
+  """
+  @spec with_locale(locale, (() -> term)) :: term
+  def with_locale(locale, fun) do
+    previous_locale = Gettext.locale
+    Gettext.locale(locale)
+    res = fun.()
+    Gettext.locale(previous_locale)
```
josevalim: We need an `after` before this line in case the function raises. :)

josevalim: If we use an after, no need for `res`.

PR #33, opened 2015-09-08
Parse empty .po(t) files correctly

### LINKED ISSUE #32: Compiler chokes on empty .po files

whatyouhide:
An empty .po(t) file is parsed as an empty list of translations. Should fix #32.


test/fixtures/empty.po
```diff
@@ -0,0 +1 @@
+
```
josevalim: What if you have a file with a bunch of new lines? :P

josevalim: Or empty spaces?

PR #35, opened 2015-09-09
Merger

whatyouhide:
This PR implements the `mix gettext.merge` task.


lib/mix/tasks/gettext.merge.ex
```diff
@@ -0,0 +1,77 @@
...
+
+  @shortdoc "Merges .pot files into existing .po files"
+
+  @doc """
+  """
+  def run(_args, priv_dir \\ "priv") do
+    _ = Mix.Project.get!
+
+    written_files =
+      priv_dir
```
josevalim: It feels we are tackling this wrong? We should rather look for the .pot files and, every time we find one, we look into the locale directories in order to perform merge? Would this help clean up some of the logic?

PR #37, opened 2015-09-10
New mix tasks

whatyouhide:
As discussed with @josevalim, `mix gettext.extract` and `mix gettext.merge` now use completely different "merging" functions; also, the behaviour of `gettext.merge` has been largely revised, and now goes something like:

```
$ mix gettext.merge PO_FILE POT_FILE

    Both files must exist

$ mix gettext.merge DIR --locale en

    Gets all POTs in DIR and sync with POs in DIR/LOCALE/LC_MESSAGES

    * If POT exists, but not PO -> create PO, fill in a couple headers
    * If PO exists, but not POT -> warn PO has no match

$ mix gettext.merge DIR

    Gets all POTs in DIR and sync with all locales in DIR/LOCALE
```


lib/gettext/extractor.ex
```diff
@@ -124,11 +129,55 @@ defmodule Gettext.Extractor do
...
+    # with the translations that only appear in `new`.
+    unique_new = Enum.reject(new.translations, &PO.Translations.find(existing.translations, &1))
+
+    %PO{translations: old_and_merged ++ unique_new, headers: existing.headers}
+  end
+
+  defp merge_translations(%Translation{} = old, %Translation{comments: []} = new) do
+    %Translation{
+      msgid: old.msgid,
+      msgstr: old.msgstr,
```
josevalim: The message strings are always empty strings. Maybe we should enforce that here and below?

lib/mix/tasks/gettext.merge.ex
```diff
@@ -2,76 +2,120 @@ defmodule Mix.Tasks.Gettext.Merge do
...
-      |> Enum.map(&write_file/1)
-
-    Enum.each written_files, fn path ->
-      Mix.shell.info "Wrote #{path}"
+    case OptionParser.parse(args, strict: [locale: :string]) do
+      {opts, [arg1, arg2], _} ->
+        run_with_two_args(arg1, arg2, opts)
+      {opts, [arg], _} ->
+        run_with_one_arg(arg, opts)
+      {_, args, _} when length(args) > 2 ->
```
josevalim: And what if there are no arguments? :)

whatyouhide: Fair point, I think I forgot to finish writing that `case` statement at some point :\

lib/mix/tasks/gettext.merge.ex
```diff
@@ -2,76 +2,120 @@ defmodule Mix.Tasks.Gettext.Merge do
...
+        Mix.raise "Too many arguments for the gettext.merge task. " <>
+                  "Use `mix help gettext.merge` to see the usage of this task."
     end
+  end
 
-    if written_files == [], do: :noop, else: :ok
+  defp run_with_two_args(arg1, arg2, []) do
+    if Path.extname(arg1) == ".po" and Path.extname(arg2) == ".pot" do
```
josevalim: The second argument can be a .po file according to msgmerge docs. So maybe we should check only the first argument to avoid a .pot file being given in the wrong order?

whatyouhide:
@josevalim I pushed a bunch of new commits addressing your comments and I think we're pretty much good to go here. I'm not particularly happy with the code as I find it messy in some places (eg, the bounds between computation and IO are not very clear in `gettext.merge` while they're super clear and clean in `gettext.extract` - guess which one you laid hands on? :D) but we can refine it even after a  0.1 release as the public API looks enough solid to me.

We're still missing docs for the tasks and some workflow docs but we can discuss them after this PR.

Wdyt?

lib/mix/tasks/gettext.extract.ex
```diff
@@ -11,7 +11,7 @@ defmodule Mix.Tasks.Gettext.Extract do
     Gettext.Extractor.setup
     force_compile
 
-    for {path, binary} <- Gettext.Extractor.dump_pot do
+    for {path, binary} <- Gettext.Extractor.pot_files do
```
josevalim: If the function is a noun, it should not shut down the extractor agent. Maybe we should provide Gettext.Extractor.teardown and call it at the end of the task.

lib/mix/tasks/gettext.merge.ex
```diff
@@ -2,76 +2,123 @@ defmodule Mix.Tasks.Gettext.Merge do
...
-    if written_files == [], do: :noop, else: :ok
+  defp run_with_two_args(arg1, arg2, []) do
+    if Path.extname(arg1) == ".po" and Path.extname(arg2) in [".po", ".pot"] do
+      ensure_file_exists!(arg1)
+      ensure_file_exists!(arg2)
+      {path, contents} = merge_po_with_pot(arg1, arg2)
+      File.write!(path, contents)
+      Mix.shell.info "Wrote #{path}"
+    else
+      Mix.raise "Arguments must be a PO file and a POT file"
```
josevalim: "and a PO/POT file"

lib/mix/tasks/gettext.extract.ex
whatyouhide: @josevalim makes sense. Personally, I like the api with `pot_files/0` and separate setup/teardown functions better (btw, it's implemented in a61db99 if you wanna have a look), because `pot_files/0` doesn't actually do anything, it only returns values right? So to me it makes sense to move setup/teardown out of it. Let me know what you think!

whatyouhide:
@josevalim fixed the stuff you addressed in the comments :)

josevalim:
Ship it!

-- 

PR #39, opened 2015-09-11
Add support for the --merge opt in gettext.extract

### LINKED ISSUE #38: Add mix gettext.extract --merge
The idea is that:

  1. The extractor layer will be able to track if a given .pot file effectively change or not

  2. If any `.pot` file changed for a given directory, for example "priv/gettext", we will call `Mix.Tasks.Gettext.Merge.run ["priv/gettext"]` once all .pot files for that directory are written

This way we can perform merge efficiently too.

whatyouhide:
This PR adds support for the `--merge` option in `gettext.extract`. @josevalim documented the option a couple of commits ago.

The implementation turned out to be pretty straightforward, I'm quite happy with the results.

Closes #38.

### DIFF
--- a/lib/mix/tasks/gettext.extract.ex
+++ b/lib/mix/tasks/gettext.extract.ex
@@ -72,8 +72,8 @@ defmodule Mix.Tasks.Gettext.Extract do
 
   defp run_merge(changed_pot_files) do
     changed_pot_files
-    |> Enum.group_by(&Path.dirname/1)
-    |> Map.keys
+    |> Enum.map(&Path.dirname/1)
+    |> Enum.uniq
     |> Enum.each(fn priv -> Mix.Task.run("gettext.merge", [priv]) end)
   end
 end
--- a/lib/mix/tasks/gettext.extract.ex
+++ b/lib/mix/tasks/gettext.extract.ex
@@ -16,16 +16,19 @@ defmodule Mix.Tasks.Gettext.Extract do
 
   """
   def run(args) do
-    changed =
-      extract
-      |> write_changed_files()
-      |> print_changed_files()
+    pot_files = extract()
+
+    for {path, contents} <- pot_files do
+      File.mkdir_p!(Path.dirname(path))
+      File.write!(path, contents)
+      Mix.shell.info "Extracted #{Path.relative_to_cwd(path)}"
+    end
 
     case args do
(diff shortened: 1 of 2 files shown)


lib/mix/tasks/gettext.extract.ex
```diff
@@ -37,4 +50,30 @@ defmodule Mix.Tasks.Gettext.Extract do
...
+  defp print_changed_files(files) do
+    Enum.each files, fn f -> Mix.shell.info("Extracted #{Path.relative_to_cwd(f)}") end
+    files
+  end
+
+  defp run_merge(changed_pot_files) do
+    changed_pot_files
+    |> Enum.map(&Path.dirname/1)
+    |> Enum.uniq
+    |> Enum.each(fn priv -> Mix.Task.run("gettext.merge", [priv]) end)
```
josevalim: We can't use `Mix.Task.run`. It will run the task only *once*. Call the fun function in it directly.

lib/mix/tasks/gettext.extract.ex
```diff
@@ -37,4 +50,30 @@ defmodule Mix.Tasks.Gettext.Extract do
   defp make_old_if_exists(path) do
     :file.change_time(path, {{2000, 1, 1}, {0, 0, 0}})
   end
+
+  defp write_changed_files(pot_files) do
+    Enum.flat_map pot_files, fn {path, contents} ->
+      File.mkdir_p! Path.dirname(path)
+
+      # Write the file only if it doesn't exist or the contents wouldn't change.
+      if not File.regular?(path) or File.read!(path) != IO.iodata_to_binary(contents) do
```
josevalim: This is ok but can't the merge algorithm only return the path if it actually change? After all, the merge algorithm is capable of finding this out... or would it be too complicated?

josevalim:
```elixir
contents = IO.iodata_to_binary(contents)
case File.read(path) do
  {:ok, ^contents} -> []
  _ ->
    File.write!(path, contents)
    [path]
end
```

whatyouhide:
@josevalim pushed a new commit and yes, you were right, we can return only the changed files in the merging algorithm, no need to do that in the task.

josevalim:
:shipit:

PR #41, opened 2015-09-15
Add the :fuzzy field to translations

whatyouhide:
The `:fuzzy` field of a translation is populated (when a translation is parsed) if there is a "fuzzy" flag in one of the flag comments (the ones starting with `#,`) for said translation. The `#,` comments are kept in the `:comments` field of a translation, similarly to how `:references` works.

@josevalim just wanted to check what you think about adding `:fuzzy`. The alternative would be to add a `:flags` field with all the flags (and maybe provide something like `Gettext.PO.Translations.fuzzy?/1`).

### DIFF
--- a/lib/gettext/po/parser.ex
+++ b/lib/gettext/po/parser.ex
@@ -31,9 +31,9 @@ defmodule Gettext.PO.Parser do
   end
 
   defp to_struct({:translation, translation}),
-    do: struct(Translation, translation) |> extract_references()
+    do: struct(Translation, translation) |> extract_references() |> set_fuzziness()
   defp to_struct({:plural_translation, translation}),
-    do: struct(PluralTranslation, translation) |> extract_references()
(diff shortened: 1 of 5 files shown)


josevalim:
We are not going to use this flag for anything, will we? I mean, we will write it down, but we will never read it for anything? If that is the case, I would use a `:flags` field and keep it a MapSet.

whatyouhide:
Yah, makes sense. We're only going to mark translations as fuzzy, so we can just add `"fuzzy"` to that MapSet and dump the MapSet with `set |> Set.to_list() |> Enum.sort` (the idea is always to change the dumped output as little as possible for VCS).

Ok, closing this PR and pushing directly to `master`, alright?

josevalim:
Sure. Just keep in mind you don't need to convert the set to a list before calling Enum.sort.

PR #42, opened 2015-09-20
Fuzzy translations

whatyouhide:
Hey @josevalim, can you have a look at this when you have some time? It's just the foundations of how fuzzy matching would work. It's very simple, but not configurable (I guess we should pass things like `--fuzzy`/`--no-fuzzy` and other options to `gettext.merge`).

Wdyt?

### DIFF
--- a/lib/gettext/merger.ex
+++ b/lib/gettext/merger.ex
@@ -5,6 +5,8 @@ defmodule Gettext.Merger do
   alias Gettext.PO.Translation
   alias Gettext.PO.PluralTranslation
 
+  @min_jaro_distance 0.8
+
   @doc """
   Merges a PO file with a POT file given their paths.
 
@@ -53,17 +55,45 @@ defmodule Gettext.Merger do
   end
 
   defp merge_translations(old, new) do
-    merged = Enum.flat_map old, fn(t) ->
+    {new, merged, obsolete} = merge_matching_translations(old, new)
+    new = match_fuzzy_translations(new, obsolete)
+    merged ++ new
+  end
+
+  defp merge_matching_translations(old, new) do
+    {new, merged, obsolete} = Enum.reduce old, {new, [], []}, fn t, {new, merged, obsolete} ->
       if same = PO.Translations.find(new, t) do
-        [merge_two_translations(t, same)]
+        {List.delete(new, same), [merge_two_translations(t, same)|merged], obsolete}
       else
-        []
+        {new, merged, [t|obsolete]}
       end
     end
 
-    new = Enum.reject(new, &PO.Translations.find(old, &1))
+    {new, Enum.reverse(merged), Enum.reverse(obsolete)}
+  end
 
-    merged ++ new
+  defp match_fuzzy_translations(new, obsolete) do
+    Enum.map new, fn(t) ->
+      if match = find_fuzzy_match(t, obsolete) do
+        %{t | msgstr: match.msgstr} |> PO.Translations.make_fuzzy
+      else
+        t
+      end
+    end
+  end
+
+  defp find_fuzzy_match(target, translations) do
+    candidates =
+      translations
+      |> Enum.map(fn(t) -> {t, PO.Translations.jaro_distance(t, target)} end)
+      |> Enum.filter(fn({_, jaro_distance}) -> jaro_distance >= @min_jaro_distance end)
+
+    if candidates == [] do
+      nil
+    else
+      {t, _jaro_distance} = Enum.max_by(candidates, fn({_, jaro_distance}) -> jaro_distance end)
+      t
+    end
   end
 
   defp merge_two_translations(%Translation{} = old, %Translation{} = new) do
--- a/lib/gettext/merger.ex
+++ b/lib/gettext/merger.ex
@@ -14,9 +14,9 @@ defmodule Gettext.Merger do
   This function returns the contents (as iodata) of the merged file, which will
   be written to a PO file.
   """
-  @spec merge_files(Path.t, Path.t) :: iodata
-  def merge_files(po_file, pot_file) do
-    merge(PO.parse_file!(po_file), PO.parse_file!(pot_file)) |> PO.dump
+  @spec merge_files(Path.t, Path.t, Keyword.t) :: iodata
+  def merge_files(po_file, pot_file, opts) do
+    merge(PO.parse_file!(po_file), PO.parse_file!(pot_file), opts) |> PO.dump
   end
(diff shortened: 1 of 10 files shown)


lib/gettext/po/translations.ex
```diff
@@ -91,4 +91,27 @@ defmodule Gettext.PO.Translations do
...
+    %{t | flags: MapSet.put(flags, "fuzzy")}
+  end
+
+  @doc """
+  Returns the Jaro distance between the msgids of the given translations.
+
+  If the translations are not of the same type (e.g., a translation and a plural
+  translation), `-1` is returned.
+  """
+  @spec jaro_distance(Gettext.PO.translation, Gettext.PO.translation) :: 0..1 | -1
```
josevalim: I don't see where `-1` would come from.

whatyouhide: Whooops, documented the spec but forgot to implement that. :)

lib/gettext/merger.ex
```diff
@@ -53,17 +55,45 @@ defmodule Gettext.Merger do
   end
 
   defp merge_translations(old, new) do
```
josevalim: I feel this is not super efficient because we are going through everything twice. We are calling calling `IO.iodata_to_binary` many times. Here is what I would try:

1. Traverse all new translations and put them in a HashDict where the PO.Translations.key() is the key and the translation is the value. Finding a matching key now becomes a dict lookup.

2. Then, if no match is found, we will ask to find the fuzzy match right there. Because we already have the keys in the "new" HashDict, we can use it and avoid more IO.iodata_to_binary calls

3. If nothing, it is obsolete

This is the first step to clean this and make this faster. Once we are done, let's discuss how to make the fuzzy stuff configurable. :D

Great job!

whatyouhide: Ok so I've given this some thought and there may be a bunch of problems (the implementation now is wrong as well btw!):

-  if we convert `new` to an HashDict, we lose ordering. If we ran the merge algorithm twice on the same input, the new translations (nor merged neither fuzzy-merged) could end up in different orders and we could get different outputs, which is not nice (hey there, VCS).
- if we ask for fuzzy matches during the first pass (whenever we don't find an exact match), we could end up with false positives: say we have translation `foo` and `foos` in the old translations, and `foos` in the new translations: if we pass `foo` first, it will probably fuzzy-match on `foos` and later on `foos` won't (because we have to take out `foos` from the new translations now that it has a match).

You're completely right about too many calls to `IO.iodata_to_binary/1` but I would solve that before doing anything else; we just map over both `new` and `old` to end up with `{key, translation}` in both of them.

I have a doubt though: when we find a fuzzy match for an old translation in the list of new translations, should we take the match out of the list of new translations? If I have `foo1` and `foo2` in the old translations and they both match on `foo`, should they both be marked as fuzzy matches?

whatyouhide:
@josevalim pushed a couple of new commits and left a bunch of comments around the code. Let me know what you think :)

josevalim:
This looks great. Now we just need how to make the fuzzy configurable as well as the threshold. We can probably do something like this:

    match(old, new, fuzzy_matcher)

Where fuzzy match receives two translations and returns `{:match, pos_integer}` and `:nomatch`. Then we could call it like this:

    match(old, new, Gettext.Fuzzy.matcher(0.8))

The matcher will follow the anonymous function signature declared above and will return `{:match, _}` only if it is above the threshold.

Thoughts?

whatyouhide:
Got it, will start working on it right now :)

whatyouhide:
@josevalim implemented `Gettext.Fuzzy.matcher/1`. Wdty?

josevalim:
Beautiful. Now let's make use of it in gettext.merge, provide the proper options, and ship it! :D

whatyouhide:
@josevalim pushed the `--fuzzy` and `--fuzzy-threshold` options, wdyt about the API?

lib/mix/tasks/gettext.merge.ex
```diff
@@ -23,14 +23,29 @@ defmodule Mix.Tasks.Gettext.Merge do
...
+
   alias Gettext.Merger
 
   def run(args) do
     _ = Mix.Project.get!
 
-    case OptionParser.parse(args, strict: [locale: :string]) do
+    parse_switches = [locale: :string, fuzzy: :boolean, fuzzy_threshold: :float]
+    case OptionParser.parse(args, strict: parse_switches) do
       {opts, [arg1, arg2], _} ->
```
josevalim: The third element of the tuple are option parsing errors. Maybe we should check they are not empty and print them somehow.

josevalim: Maybe something like this:

     {_, _, [_|_] = errors} ->
      for {key, _} <- errors, do: Mix.shell.error "#{key} is invalid"
      Mix.raise "mix gettext.merge aborted."

lib/mix/tasks/gettext.merge.ex
```diff
@@ -23,14 +23,29 @@ defmodule Mix.Tasks.Gettext.Merge do
...
+  @default_merging_opts fuzzy: true, fuzzy_threshold: 0.8
+
   alias Gettext.Merger
 
   def run(args) do
     _ = Mix.Project.get!
 
-    case OptionParser.parse(args, strict: [locale: :string]) do
+    parse_switches = [locale: :string, fuzzy: :boolean, fuzzy_threshold: :float]
+    case OptionParser.parse(args, strict: parse_switches) do
```
josevalim: Don't use `:strict` because we are receiving all args from extract, so options given to extract could leak here and make it fail. `:switches` would be better.

whatyouhide: Very good call, pushed a new commit right now :). Let me know what you think!

lib/mix/tasks/gettext.extract.ex
```diff
@@ -19,6 +19,12 @@ defmodule Mix.Tasks.Gettext.Extract do
 
       mix gettext.extract --merge
 
+  The following options are forwarded to the `gettext.merge` task, which is
+  called internally by this task:
+
+    * `--fuzzy`
```
josevalim: `--no-fuzzy`. I would actually reword the paragraph to say that it supports the same option as `gettext.merge` in those cases and just show an example. :)

whatyouhide:
@josevalim pushed the fixes to the docs, I'm merging this PR. I'll polish the docs a little bit and then we can release 0.6.0 :tada:

Issue #43, opened 2015-09-20
gettext.merge/extract can use Task.async/await


josevalim:
As long as they are writing to different files, we are good to go.

PR #44, opened 2015-09-21
Parallelizing tasks

whatyouhide:
As per #43. This PR just makes the `gettext.extract` and `gettext.merge` tasks more parallel using `Task.async/1` and `Task.await/1`. Should be nice and easy.


josevalim:

PR #45, opened 2015-10-03
Move :fuzzy_threshold out of :gettext's env

whatyouhide:
@josevalim we store the default for `:fuzzy_threshold` in `:gettext`'s `:env` but we don't load gettext here in a mix task I guess? We don't have the env anyway. FWIW, It also works if I leave the option in the `:env` and call `Application.load(:gettext)` before fetching it, not sure which one is better here (this looks simpler though).

### DIFF
--- a/lib/mix/tasks/gettext.merge.ex
+++ b/lib/mix/tasks/gettext.merge.ex
@@ -83,6 +83,8 @@ defmodule Mix.Tasks.Gettext.Merge do
 
   """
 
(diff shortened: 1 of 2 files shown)


josevalim:
We would need to start the gettext app. Which you can easily do. But this
looks good to go as well!

-- 


whatyouhide:
I'll merge this for simplicity right now so that I can release 0.6.1, if we'll need to start the application for something else than we can change this back as well!

PR #47, opened 2015-10-17
References

### LINKED ISSUE #46: Cannot read multiple #: references on one line
When I use Poedit it saves the .po file like this:

```
#: web/views/dashboard_view.ex:5 web/views/dashboard_view.ex:10
#: web/views/dashboard_view.ex:20
msgid "Memory"
msgstr "Minne"
```

with multiple source code references on same line (there is a wrap-at preference setting in poedit).

This makes gettext choke:

````
== Compilation error on file lib/brando/gettext.ex ==
** (MatchError) no match of right hand side value: ["web/views/dashboard_view.ex", "5 web/views/dashboard_view.ex", "10"]
    lib/gettext/po/parser.ex:53: Gettext.PO.Parser.parse_reference/1
    lib/gettext/po/parser.ex:47: anonymous fn/2 in Gettext.PO.Parser.extract_references/1
    (elixir) lib/enum.ex:1385: Enum."-reduce/3-lists^foldl/2-0-"/3
    lib/gettext/po/parser.ex:45: Gettext.PO.Parser.extract_references/1
    lib/gettext/po/parser.ex:34: Gettext.PO.Parser.to_struct/1
    (elixir) lib/enum.ex:1043: anonymous fn/3 in Enum.map/2
    (elixir) lib/enum.ex:1385: Enum."-reduce/3-lists^foldl/2-0-"/3
    (elixir) lib/enum.ex:1043: Enum.map/2
```

This looks to be a valid way to print references, at least according to the GNU gettext manual:

```
#: src/msgcmp.c:338 src/po-lex.c:699
#, c-format
msgid "found %d fatal error"
msgid_plural "found %d fatal errors"
msgstr[0] "s'ha trobat %d error fatal"
msgstr[1] "s'han trobat %d errors fatals"
```

whatyouhide:
@josevalim I'm not proud of the implementation, but the feature works: parsing multiple references on the same line and dumping PO with all the references joined in one line and then wrapped.

On IRC the problem of spaces in file names arrived. I looked at `xgettext` and with this:

```
xgettext "foo bar.c" foo.c
```

you get

```po
# foo bar.c:1 foo.c:1
msgid "foo"
msgstr "bar"
```

So I pretty much decided to use line numbers to separate filenames. So after a line number, everything up to the next line number is the file name. Let me know what you think!

This would close #46.

/cc @tmjoen

### DIFF
--- a/lib/gettext/po.ex
+++ b/lib/gettext/po.ex
@@ -22,6 +22,8 @@ defmodule Gettext.PO do
     file: Path.t,
   }
 
+  @reference_wrapping_column 80
+
   @doc """
   Parses a string into a list of translations.
 
@@ -238,10 +240,28 @@ defmodule Gettext.PO do
     Enum.map(comments, &[&1, ?\n])
   end
 
+  defp dump_references([]) do
+    ""
+  end
+
   defp dump_references(references) do
-    for {file, line} <- references do
-      "#: #{file}:#{line}\n"
+    # This function outputs a bunch of #: comments with as many references on
+    # each line as there can be under @reference_wrapping_column columns.
+    # In the flat_map_reduce, we insert newlines and comment delimiters (#:)
+    # when references wrap, building an iolist. At the end, we need to insert
+    # the first comment delimiter and the last newline.
+    wrapping_column = @reference_wrapping_column - String.length("#:")
+    {iolist, _} = Enum.flat_map_reduce references, 0, fn({file, line}, line_length) ->
+      ref = " #{file}:#{line}"
+      ref_length = String.length(ref)
+      if (ref_length + line_length) > wrapping_column do
+        {[?\n, "#:", ref], ref_length}
+      else
+        {[ref], line_length + ref_length}
+      end
     end
+
+    ["#:", iolist, ?\n]
   end
 
   defp dump_flags(flags) do
--- a/lib/gettext/po/parser.ex
+++ b/lib/gettext/po/parser.ex
@@ -42,16 +42,28 @@ defmodule Gettext.PO.Parser do
   defp extract_references(%{__struct__: _, comments: comments} = translation) do
     {reference_comments, other_comments} = Enum.partition(comments, &match?("#:" <> _, &1))
     references =
-      for "#:" <> contents <- reference_comments,
-        (contents = String.strip(contents)) != "",
-        do: parse_reference(contents)
+      reference_comments
+      |> Enum.reject(fn("#:" <> comm) -> String.strip(comm) == "" end)
+      |> Enum.flat_map(&parse_references/1)
 
     %{translation | references: references, comments: other_comments}
   end
 
-  defp parse_reference(ref) do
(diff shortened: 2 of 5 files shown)


tmjoen:
Tried it now, works great. Grazie mille!

lib/gettext/po.ex
```diff
@@ -238,10 +240,28 @@ defmodule Gettext.PO do
...
+
   defp dump_references(references) do
-    for {file, line} <- references do
-      "#: #{file}:#{line}\n"
+    # This function outputs a bunch of #: comments with as many references on
+    # each line as there can be under @reference_wrapping_column columns.
+    # In the flat_map_reduce, we insert newlines and comment delimiters (#:)
+    # when references wrap, building an iolist. At the end, we need to insert
+    # the first comment delimiter and the last newline.
+    wrapping_column = @reference_wrapping_column - String.length("#:")
```
josevalim: `String.length("#:")` can be moved to a module attribute.

lib/gettext/po.ex
```diff
@@ -238,10 +240,28 @@ defmodule Gettext.PO do
...
+    {iolist, _} = Enum.flat_map_reduce references, 0, fn({file, line}, line_length) ->
```
josevalim: Just an idea, this may be simpler if you first group the times that fit a single line and then generate the lines. But maybe not. :)

lib/gettext/po/parser.ex
```diff
@@ -42,16 +42,28 @@ defmodule Gettext.PO.Parser do
...
     %{translation | references: references, comments: other_comments}
   end
 
-  defp parse_reference(ref) do
-    [file, line] = String.split(ref, ":")
-    {file, String.to_integer(line)}
+  defp parse_references("#:" <> comment) do
+    comment
+    |> String.strip()
+    |> String.split(":")                      # we remain with "21 foo.ex"
```
josevalim: Love the comments.

josevalim:

lib/gettext/po.ex
whatyouhide: @josevalim sorry for the late response, was out of town for a few days :) I'm not sure what you mean here, something like this?

```elixir
references = Enum.map(references, fn({file, line}) -> " #{file}:#{line}" end)

{translations_per_line, _} = Enum.reduce references, {[0], 0}, fn(ref, {[latest|rest], line_length}) ->
  ref_length = String.length(ref)

  if (ref_length + line_length) > wrapping_column do
    {[1, latest|rest], ref_length}
  else
    {[latest + 1|rest], line_length + ref_length}
  end
end

{lines, []} = Enum.flat_map_reduce Enum.reverse(translations_per_line), references, fn n, acc ->
  {line, acc} = Enum.split(acc, n)
  {["#:", line, ?\n], acc}
end

lines
```

josevalim: Yeah, it doesn't help at all I guess. :(

josevalim: :shipit: as is!

Issue #48, opened 2015-10-19
Consider having default locale set is a must,


lexmag:
otherwise we need a more informative error message for a following case:
```elixir
iex(1)> Gettext.locale
nil
iex(2)> Gettext.with_locale("it", fn() -> end)
** (ArgumentError) locale/1 only accepts binary locales
    lib/gettext.ex:432: Gettext.locale/1
    lib/gettext.ex:568: Gettext.with_locale/2
iex(3)> Gettext.locale
"it"
```
As you see without `:default_locale` env var it fails to reset locale back.

josevalim:
Yes, `Gettext.locale` always need to return something. It may just default to "en".

josevalim:
I am also wondering if the locale configuration should really be in `Gettext.locale` or if it should be per backend.

lexmag:
> locale configuration should really be in Gettext.locale or if it should be per backend
It feels more intuitive to have it per backend, and more flexible.

whatyouhide:
I'm :+1: on making the locale per-backend. We can store it under `MyApp.Gettext` in the process dict and increment the arity of `locale/0-1` to `locale(backend)` and `locale(backend, locale)`. I'm not sure how we could configure it. We could use the Mix config for each backend:

```elixir
config :my_app, MyApp.Gettext, default_locale: "pt_BR"
```

and also default to `"en"` if there's no config (which is the subject of the issue).

What do you think?

josevalim:
:+1: for everything. Will we also have `MyApp.Gettext.locale/0/1`?

lexmag:
That's sounds wonderful.

whatyouhide:
@josevalim I wouldn't provide both `Gettext.locale/1-2` and `MyApp.Gettext.locale/0-1` because they would do the exact same thing (but we would still have to test/document all of them). Also, we'd need to introduce `MyApp.Gettext.with_locale/2` as well. Wdyt?

lexmag:
I do see the need for `MyApp.Gettext.with_locale/2`, not sure if we need `MyApp.Gettext.locale/0-1` though.

josevalim:
Ok, we don't need to add any of them. `Gettext.get_locale(backend)` & `Gettext.put_locale(backend, ...)` would be my votes. I think it would be nice to change the name so it is clearer now with multiple args and so we effectively break the users code, otherwise it can accidentally pass.

whatyouhide:
[...]

PR #49, opened 2015-10-21
Per-backend locale

whatyouhide:
Implementation of what was discussed in #48. Wdyt? /cc @lexmag @josevalim

### DIFF
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -85,25 +85,26 @@ defmodule Gettext do
   ## Locale
 
   At runtime, all gettext-related functions and macros that do not explicitely
-  take a locale as an argument read the locale from `Gettext.locale/0`. The
-  locale can be set with `Gettext.locale/1`. Locales are expressed as strings
-  (like `"en"` or `"fr"`); they can be arbitrary strings as long as they match a
-  directory name.
+  take a locale as an argument read the locale from `Gettext.get_locale/1`. The
+  locale can be set with `Gettext.put_locale/2`. Locales are expressed as
+  strings (like `"en"` or `"fr"`); they can be arbitrary strings as long as they
+  match a directory name.
 
-  Gettext stores the locale **per-process** (in the process dictionary). This
-  means that `Gettext.locale/1` must be called in every new process in order to
-  have the right locale available in that process. Pay attention to this
-  behaviour, since not setting the locale with `Gettext.locale/1` *will not*
-  result in any errors when `Gettext.locale/0` is called; the default locale
-  will be returned instead.
+  Gettext stores the locale **per-process** (in the process dictionary) and per
+  Gettext module. This means that `Gettext.put_locale/2` must be called in every
+  new process in order to have the right locale available in that process. Pay
+  attention to this behaviour, since not setting the locale with
+  `Gettext.put_locale/2` *will not* result in any errors when `Gettext.get_locale/1`
+  is called; the default locale will be returned instead.
 
   ### Default locale
 
-  The default Gettext locale is `"en"`. The value of the default locale can be
-  modified in the configuration for the `:gettext` application. For example, in
-  the `config/config.exs` file of the `my_app` application:
+  The default Gettext locale is `"en"`. The value of the default locale for a
+  given Gettext module can be set in the configuration for the `:otp_app` of
+  that Gettext module. For example, in the `config/config.exs` file of the
+  `my_app` application:
 
-      config :gettext, default_locale: "fr"
+      config :my_app, MyApp.Gettext, default_locale: "fr"
 
   ## Gettext API
 
@@ -145,7 +146,7 @@ defmodule Gettext do
 
   These are all valid uses of the gettext macros:
 
-      Gettext.locale "it"
+      Gettext.put_locale MyApp.Gettext, "it"
 
       MyApp.Gettext.gettext "Hello world"
       #=> "Ciao mondo"
@@ -176,7 +177,7 @@ defmodule Gettext do
         use Gettext, otp_app: :my_app
       end
 
-      Gettext.locale "pt_BR"
+      Gettext.put_locale MyApp.Gettext, "pt_BR"
 
       msgid = "Hello world"
       Gettext.gettext(MyApp.Gettext, msgid)
@@ -216,7 +217,7 @@ defmodule Gettext do
 
   interpolation can be done like follows:
 
-      Gettext.locale "it"
+      Gettext.put_locale MyApp.Gettext, "it"
       MyApp.Gettext.gettext "Hello, %{name}!", name: "Meg"
       #=> "Ciao, Meg!"
 
@@ -246,7 +247,7 @@ defmodule Gettext do
 
   the `ngettext` macro can be used like this:
 
-      Gettext.locale "it"
+      Gettext.put_locale MyApp.Gettext, "it"
       MyApp.Gettext.ngettext "One error", "%{count} errors", 3
       #=> "3 errori"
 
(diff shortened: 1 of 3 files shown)


lib/gettext.ex
```diff
@@ -386,50 +385,52 @@ defmodule Gettext do
...
-    if locale = Process.get(__MODULE__) do
+  @spec get_locale(backend) :: locale
+  def get_locale(backend) do
+    if locale = Process.get(backend) do
       locale
     else
-      default_locale = Application.get_env(:gettext, :default_locale)
-      Process.put(__MODULE__, default_locale)
+      backend_config = Application.get_env(backend.__gettext__(:otp_app), backend, [])
+      default_locale = backend_config[:default_locale] || "en"
```
lexmag: I think `Keyword.get/3` would be better here, as we don't silently rewrite `default_locale: nil` case.

whatyouhide: Makes sense, pushed the change you suggested.

lib/gettext.ex
```diff
@@ -246,7 +247,7 @@ defmodule Gettext do
 
   the `ngettext` macro can be used like this:
 
-      Gettext.locale "it"
+      Gettext.locale(MyApp.Gettext, "it")
```
lexmag: `put_locale`

lib/gettext.ex
```diff
@@ -281,7 +282,7 @@ defmodule Gettext do
 
   For example:
 
-      Gettext.locale "foo"
+      Gettext.locale(MyApp.Gettext, "foo")
```
lexmag: `put_locale`

lexmag:
I really feel that per backend `with_locale` will be handy.
As today we expect users to `import MyApp.Gettext`, but not having `with_locale` imported may defeat the the value of such import I think.
For instance:
```elixir
  import MyApp.Gettext

  defp translate_header(locale, bindings) do
    Gettext.with_locale(MyApp.Gettext, locale, fn() ->
      gettext <some text>, bindings
    end)
  end

  defp translate_body(locale) do
    Gettext.with_locale(MyApp.Gettext, locale, fn() ->
      gettext <other text>
    end)
  end
```
Seems verbose and I believe people will end up wrapping `Gettext.with_locale(MyApp.Gettext, locale, fun)` in their own function.

whatyouhide:
Verbosity is without doubt the most negative thing about the `Gettext.*locale

josevalim:
@lexmag I don't think folks will use with_locale like that, honestly. You will usually set the locale only once, at the beginning of your requests in your plug pipeline, and that's it. No need to swap the locale for every translation, with_locale is mostly for tests.

josevalim:
:shipit:

whatyouhide:
@josevalim yes, I agree setting the locale is something you do once and getting it is not a frequent operation probably. So you're in favour of `Gettext.*locale` functions versus `MyApp.Gettext.*locale` functions?

lexmag:
@josevalim there is one real case I've encountered already – generating payload for push notifications. :smile:

lexmag:
Anyway, I always for simplicity at starting, I'll just be wrapping it for now.

lexmag:
I think I could avoid using `with_locale` in my case, the only thing that is needed then is `MyApp.Gettext.default_locale/0` – locale that can't be affected by any means.
Today it's doable with the following dance:

```elixir
defmodule MyApp.Gettext do
  use Gettext, otp_app: :my_app

  def default_locale() do
    __MODULE__.Defaults.__locale__
  end
end

defmodule MyApp.Gettext.Defaults do
  def __locale__() do
    unquote(Gettext.get_locale(MyApp.Gettext))
  end
end
```
@whatyouhide @josevalim what are your thoughts on having `default_locale/0`?

whatyouhide:
@lexmag I'm probably missing something, but couldn't you do just:

```elixir
defmodule MyApp.Gettext do
  use Gettext, otp_app: :my_app

  def default_locale do
    unquote(Gettext.get_locale(__MODULE__))
  end
end
```

Btw, still assuming I'm not missing something, I would go with `Gettext.default_locale/1` given we can always retrieve the default locale with `Application.get_env/3`.

lexmag:
@whatyouhide you can't do this without extra supporting module as `module MyApp.Gettext is not available` during `unquote`.

whatyouhide:
@lexmag I see, but we could implement it as:

```elixir
defmodule Gettext do
  def default_locale(backend) do
    Application.get_env(backend.__gettext__(:otp_app), backend)) |> Keyword.get(:default_locale, "en")
  end
end
```

right?

whatyouhide:
@lexmag or do you mean that people can mess with that using `Application.put_env/2`?

lexmag:
If we go with `Gettext.default_locale/1` function I think `Application.get_env/2` is fine (extra `ets` call though).
If one mess with `Application.put_env/2` it's honestly not our problem.

whatyouhide:
I 100% agree on that (people using `put_env/3`), that would have been my argument :). Personally I don't think the extra ets call is a problem as I don't think `default_locale/1` is a critical function. If it turns out to be, I'm more inclined to change all the `*locale` functions and move them to `MyApp.Gettext` rather than having `put_/get_/with_locale` in `Gettext` and `default_locale` in `MyApp.Gettext`. Wdyt?

lexmag:
After some tryouts of v0.7 I believe we should think of backend modules as of main basic thing to interact with locales settings, i.e. having `*locale` functions in `MyApp.Gettext` feels truly natural.

whatyouhide:
@lexmag I think I agree with that. Since they're per-module, it feels more natural to have them set by their module. Also, it's less verbose :smiley_cat: Let's hear what @josevalim thinks?

josevalim:
:+1: for being per module although I would not add with_locale to the modules. Again, it is alright to be in Gettext as I most see it as a helper and not a first class API.

whatyouhide:
Just one last doubt before moving the functions from `Gettext` to `MyApp.Gettext`: keeping them in `Gettext` has the benefit of easy documentation. Today, we don't generate docs for `*gettext` macros for example, so this is something we have to solve either way. Wdyt? /cc @josevalim @lexmag

josevalim:
Sounds alright.

lexmag:
Ping.

whatyouhide:
oh god 0.0 pong, sorry!! Will work on this tomorrow, i promise :/

On Friday, 20 November 2015, Aleksei Magusev <notifications@github.com>
wrote:

> Ping.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/elixir-lang/gettext/pull/49#issuecomment-158519837>.
>


-- 

Andrea Leopardi
an.leopardi@gmail.com

whatyouhide:
@lexmag hey there, so just a question: how would we go about documentation if we move all functions and macros from `Gettex` to `MyApp.Gettext`? I suppose we could generate them but I still like that people can look at the documentation for gettext without having to bring gettext + ex_doc in their projects :\

josevalim:
@whatyouhide one common trick is to define them as `@callback` in the Gettext module. This way, you still have a space for documenting them and you can declare that each backend implements the `Gettext` module behaviour.

josevalim:
@whatyouhide btw, what was the decision? will we add put_locale/get_locale to the backends as well?

whatyouhide:
@josevalim I think we settled on *moving* `put_/get_locale` to the backends only, removing them from `Gettext`. Can you think of any downsides in doing that? We would leave `with_locale` in Gettext though, as as you said it's meant mostly for testing.

lexmag:
@whatyouhide as José said, `@callback` definition would be the best option, thus we will have cross-references from implementation to behaviour, and decent docs for callbacks.

I think we agreed on:
* move `put_/get_locale` to the backends
* keep `with_locale` in `Gettext`
* introduce `default_locale/0` in the backends

whatyouhide:
Alright, sounds like a plan. Working on it now :)

josevalim:
I feel like we should keel `put_/get_locale` in `Gettext` only because it has the implementation of everything else. Also, I wouldn't introduce `default_locale/0` only because you can easily read it from the app config. Since all of this is generated, it is nice to keep those as small as possible.

lexmag:
Okay, let's wait with these changes then, until more feedback comes. :smile:

PR #53, opened 2015-10-25
Fix docs for get_locale and put_locale

padde:
The backend parameter was missing in some examples.


whatyouhide:
Hey @padde, thanks for spotting this :) <3

PR #56, opened 2015-11-21
Fallback to default clauses when msgstrs are empty

### LINKED ISSUE #55: gettext.extract --merge does not preserve default msgstr's
I may be missing something, but when running `mix gettext.extract --merge` for the first time, I find all the message strings are blank. This results in all templates returning blank strings until I go fill in the translations. It seems like a better approach would be to have the msgid copied over to the msgstr. Thoughts?

Example of generated .po

```
#: web/views/page_view.ex:3
msgid "A productive web framework that does not compromise speed and maintainability."
msgstr ""

#: web/views/layout_view.ex:19
msgid "Get Started"
msgstr ""
```

Example of what I'd like to be generated:

```
#: web/views/page_view.ex:3
msgid "A productive web framework that does not compromise speed and maintainability."
msgstr "A productive web framework that does not compromise speed and maintainability."

#: web/views/layout_view.ex:19
msgid "Get Started"
msgstr "Get Started"
```

whatyouhide:
Should close #55.

Before this PR, empty msgstrs were treated like any other msgstr. This resulted in a nasty behaviour when generating new PO files (from POT files) with e.g. `mix gettext.merge`, because empty
msgstrs were being returned.

This PR changes the behaviour of the `Gettext.Compiler`, which now doesn't generate a function clause for translations with an empty msgstr at all. This means such translations will fallback to the dynamic clauses of `lgettext` and `lngettext`, which just return the msgid (after interpolation) in a `{:default, _}` tuple. Plural translations are treated this way if they have one or more empty msgstrs.

/cc @chrismccord and @endersstocker

### DIFF
--- a/lib/gettext/compiler.ex
+++ b/lib/gettext/compiler.ex
@@ -146,7 +146,8 @@ defmodule Gettext.Compiler do
   end
 
   # `acc` is a list of already compiled translation, i.e., of quoted function
-  # definitions.
+  # definitions. Here, we prepend a quoted function definition to that list for
+  # each translation in the given PO file.
   defp compile_po_file(path, acc) do
     {locale, domain} = locale_and_domain_from_path(path)
     %PO{translations: translations} = PO.parse_file!(path)
@@ -166,9 +167,15 @@ defmodule Gettext.Compiler do
     msgid  = IO.iodata_to_binary(t.msgid)
     msgstr = IO.iodata_to_binary(t.msgstr)
 
-    quote do
-      def lgettext(unquote(locale), unquote(domain), unquote(msgid), var!(bindings)) do
-        unquote(compile_interpolation(msgstr))
+    # Only actually generate this function clause if the msgstr is not empty. If
+    # it's empty, not generating this clause (by returning `nil` from this `if`)
+    # means that the dynamic clause will be executed, returning `{:default,
+    # msgid}` (with interpolation and so on).
+    if msgstr != "" do
+      quote do
+        def lgettext(unquote(locale), unquote(domain), unquote(msgid), var!(bindings)) do
+          unquote(compile_interpolation(msgstr))
+        end
       end
     end
   end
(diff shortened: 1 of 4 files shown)


lib/gettext/compiler.ex
```diff
@@ -166,9 +166,16 @@ defmodule Gettext.Compiler do
...
     msgstr = IO.iodata_to_binary(t.msgstr)
 
-    quote do
-      def lgettext(unquote(locale), unquote(domain), unquote(msgid), var!(bindings)) do
-        unquote(compile_interpolation(msgstr))
+    # Only actually generate this function clause if the msgstr is not empty. If
+    # it's empty, not generating this clause means that the dynamic clause will
+    # be executed, returning `{:default, msgid}` (with interpolation and so on).
+    if msgstr == "" do
+      nil
```
lexmag: Maybe just omit this branch?

lib/gettext/compiler.ex
```diff
@@ -177,17 +184,25 @@ defmodule Gettext.Compiler do
...
 
-    quote do
-      def lngettext(unquote(locale), unquote(domain), unquote(msgid), unquote(msgid_plural), n, bindings) do
-        plural_form    = unquote(@plural_forms).plural(unquote(locale), n)
-        var!(bindings) = Map.put(bindings, :count, n)
+    # If any of the msgstrs is empty, then we skip the generation of this
+    # function clause. The reason we do this is the same as for the
+    # `%Translation{}` clause.
+    if Enum.any?(msgstr, &match?({_, ""}, &1)) do
+      nil
```
lexmag: Same here. :)

lib/gettext/compiler.ex
whatyouhide: @lexmag I'm not a fan of implicit `nil`s for `if` actually :( I know you are (it's in the styleguide :smile:) but I think that often relying on the default `nil` is a very implicit behaviour which makes it harder to understand what the code is doing. I think here this may be even stronger, because we're relying on the fact that `nil` is a valid quoted term that in this instance means "don't generate anything"; it's already slightly explicit this way, I'm afraid omitting the `nil` branch would only make it worse :| wdyt?

lexmag: @whatyouhide I could see the value of explicit `nil` if we're interested in the returning value (e.g. assignment), but when `if` has the meaning of skipping some work I believe it is much more clear without `nil`. :bowtie:
(Personally, I unfold the diff above this change to see – maybe we are doing something with this `nil` :|).

whatyouhide: @lexmag I pushed a compromise, there's just one `if` (or `unless`) branch and the `nil` is implicit but I mention we're sometimes returning `nil` in the comments :smile: wdyt?

lexmag: Nice. :+1: 
_(however this comment feels superfluous as it it just how `if` works ¯\_(ツ)_/¯)_

whatyouhide: @lexmag I know, I know :smile: but I value prose and explicitness too much to let this go :smile_cat: jokes aside, I can even remove the `(by returning nil from this if)` if you want!

josevalim:

josevalim:
Let's ship it and see if there is anything left before we do a new release. :D

PR #59, opened 2015-11-25
Update README instructions to 0.8

stevedomin:
Just so people don't use the wrong versions in their projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 
     ```elixir
     def deps do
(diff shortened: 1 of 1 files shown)


stevedomin:
(sorry, always making the same PR :stuck_out_tongue:)

whatyouhide:
@stevedomin ahah, I'll never remember to update that. Thanks :)

PR #61, opened 2015-11-29
Better error message for non compile-time strings

whatyouhide:
@josevalim makes sense? I just reworded the error message very slightly (e.g., I say "use the functions in the Gettext module" instead of "use the Gettext module").


josevalim:
:shipit:

Issue #64, opened 2015-12-10
Add comments to generated files


josevalim:
For .pot files we must say: 

```
# msgids are often extracted from source code.
# Run "mix gettext.extract" to make them up to date.
```

For .po files we must say:

```
# msgids come from .pot files. Do not add, change or remove msgids manually, only msgstr.
# Use "mix gettext.extract --merge" or "mix gettext.merge" to merge .pot into .po.
```

We also need a mechanism to have those comments in pot files not leak into po.

PR #66, opened 2015-12-11
[WIP] Add informative comments to new POT files

whatyouhide:
@josevalim @chrismccord this is a WIP but I'd love some early feedback on the direction this is taking. For now, I only implemented the informative comment for newly generated POT files. There are quite a lot of changes for a simple feature, but we weren't distinguishing new/merged/only old POT files in the extractor, so I had to work on that.

Btw, I also thought about just checking the list of `{path, pot_contents}` tuples before writing them and adding the comment if it wasn't there, but then people wouldn't have been able to remove the comment. This way, the comment is only added to newly generated POT files (if it's removed and `gettext.extract` is run again, it's not readded).

Related to #64.


lib/gettext/extractor.ex
```diff
@@ -158,12 +158,44 @@ defmodule Gettext.Extractor do
...
+  defp dump_tagged_file({path, {tag, po}}) when tag in [:unmerged, :merged],
+    do: {path, PO.dump(po)}
+
+  defp prune_unmerged(path) do
+    merge_or_unchanged(path, %PO{})
+  end
+
+  defp new_pot_comment do
+    """
+    # This file is a PO template file. `msgid`s here are often extracted from
```
josevalim: "PO Template" (in case PO template is not obvious that it maps to POT).

josevalim: How are we guaranteeing that this is not copied to PO files? My concern is that, if we change the text, it will then end-up suddenly merging. Should we adopt a convention where comments starting with two `##` won't be copied over?

whatyouhide: @josevalim I still haven't take care of that, that's also why the WIP :). The `##` sounds good. I think right now, we're in a bit of a weird (and wrong) situation because I don't think the comments at the top of the file are copied over PO files but only if the first translation is the `msgid ""` headers translation. I have to fix that by bringing the comments of the headers translation along (which are the comments at the top of the file), and we can just not bring `##` comments over. Wdyt?

josevalim: I believe they are copied today and they would always apply to the first translation (not the empty one). That's why I think always ignoring the ones `##` would be helpful.

josevalim: For example, in Phoenix pull request, we have added things like `## From Ecto.Changeset.cast/4` to POT files but this information is useless to translators.

lib/gettext/extractor.ex
```diff
@@ -129,17 +129,17 @@ defmodule Gettext.Extractor do
...
     # extracted. If we turn pot_files into a list of {path, whatever} tuples,
     # that we can take advantage of Dict.merge/3 to find clashing paths.
-    pot_files
-    |> Enum.map(&{&1, :existing})
-    |> Enum.into(%{})
-    |> Map.merge(Enum.into(po_structs, %{}), &merge_existing_and_extracted/3)
-    |> Enum.map(&purge_unmerged_files/1)
-    |> Enum.reject(&match?({_, :unchanged}, &1))
-    |> Enum.map(fn({path, pot}) -> {path, PO.dump(pot)} end)
+    pot_files  = Enum.map(pot_files, &{&1, :existing}) |> Enum.into(%{})
```
lexmag: We have `Enum.into/3` = `Enum.map/2` + `Enum.into/2`.

whatyouhide:
Just pushed new commits; after thinking about it for some time, I guess the only place where we need to strip such `##` comments is `Gettext.Merger.new_po_file/2`, which is a function that takes an existing POT file and the filename for a new PO file and basically merges the POT file with the new PO file. This is the only place where translator comments from the POT file are brought over to the PO file; in `Gettext.Merger.merge_files/2` (which is used to "update" PO files), translator comments from the POT file are discarded (including `##` comments obviously).

Wdyt?

josevalim:
Looks great. Let's also add the .PO comment and :shipit:

whatyouhide:
Pushed a new commit that writes informative comments when generating **new** PO files. I had to change quite a lot of things because right now, we're parsing headers as a `PO.Translation` first, then ditching all the comments and metadata and only keeping the raw list of headers. This caused top-of-the-file comments (above the headers) to be discarded, which we don't want as a general rule (regardless of those comments being generated by us or handwritten by translators).

So, I couldn't come up with a better way than adding a `:top_of_the_file_comments` (ugh, bad name) field to the `%PO{}` struct; this is filled out in `PO.parse_string/1` because the parser now returns `{:ok, top_comments, headers, translations}` instead of `{:ok, headers, translations}`.

As of the last commit, when new PO files are generated the informative comment is inserted but whenever they're re-written (because of merging or by the user), it's not added again (so if the user removes it, it stays removed).

Wdyt? /cc @josevalim @lexmag

whatyouhide:
Btw, there may be some cleaning up/refactoring to do; let's just see if this looks like a good idea and then I'll clean up in case it is :)

josevalim:
Can't you make the top of the file comments to be the comments to the empty `msgid` string which is usually at the top of the file? This way you don't need to add new fields.

whatyouhide:
@josevalim we don't keep headers as a `Translation` struct, but as a list of strings (here: https://github.com/elixir-lang/gettext/blob/master/lib/gettext/po/parser.ex#L81-L89) :\

josevalim:
I see, than this looks good! :D

whatyouhide:
Ok, do I :shipit: and release 0.9?

josevalim:
Yes!

PR #68, opened 2015-12-22
Make example more illustrative

paulhenrich:
I'm at the Montevideo Elixir meetup, and we have some ideas for
improving the documentation. The idea here is to make the example
less abstract, and clarify the meaning of "3" by providing a
map with some context.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,9 +52,10 @@ import MyApp.Gettext
 gettext "Here is one string to translate"
 
 # Plural translation
-ngettext "Here is the string to translate",
-         "Here are the strings to translate",
-         3
+ngettext "The apple is ripe",
+         "The apples are ripe",
+         %{thing: 'apple', count: 3}.count
+# Returns "The apples are ripe"
 
 # Domain-based translation
 dgettext "errors", "Here is an error message to translate"


README.md
```diff
@@ -52,9 +52,11 @@ import MyApp.Gettext
 gettext "Here is one string to translate"
 
 # Plural translation
-ngettext "Here is the string to translate",
-         "Here are the strings to translate",
-         3
+ngettext "The apple is ripe",
+         "The apples are ripe",
+         %{ thing: 'apple', size: 3 }.size
```
josevalim: Cosmetic detail: no spaces after `{` and before `}`. Thank you!

lexmag: Maybe `size` –> `count`?

paulhenrich:
@lexmag @josevalim Thanks, I amended the commit with your suggested changes.

josevalim:

whatyouhide:
My 2 cents: isn't using a map and then calling `.count` on it slightly more confusing than something like this:

```elixir
count = 3
ngettext "foo",
         "foos",
         count
```

lexmag:
Agree with @whatyouhide.

paulhenrich:
I'm fine with making that change. I agree it's a bit clearer.

Issue #70, opened 2015-12-28
Terrible error message when passing a non binary to Gettext.*gettext functions


whatyouhide:
When a non binary value is passed to one of the `Gettext.*gettext` functions, the error is absolutely terrible (it goes down to `Regex.split/3`, making it super cryptic):

```
iex(3)> Gettext.dgettext MyApp.Gettext, "default", :foo
** (FunctionClauseError) no function clause matching in Regex.split/3
    (elixir) lib/regex.ex:375: Regex.split(~r/\n    (?<left>)  # Start, available through :left\n    %{         # Literal '%{'\n
    [^}]+    # One or more non-} characters\n    }          # Literal '}'\n    (?<right>) # End, available through :right\n  /x, :
foo, [on: [:left, :right], trim: true])
             lib/gettext/interpolation.ex:27: Gettext.Interpolation.to_interpolatable/1
             lib/gettext/interpolation.ex:99: Gettext.Interpolation.interpolate/2
    (my_app) lib/my_app/gettext.ex:1: MyApp.Gettext.lgettext/4
             lib/gettext.ex:480: Gettext.dgettext/4
```

PR #71, opened 2015-12-28
Fail more gracefully in Gettext.*gettext functions

whatyouhide:
@josevalim what do you think of this to solve #70?

/cc @lexmag which always has good advices :)


lexmag:
:+1: don't think we need extra clause with manual `raise Argument` as `FunctionClauseError` is good enough and always shows bad call arguments.

lexmag:
@whatyouhide I'm following Gettext and will come to the PR anyway. :wink:

josevalim:
:shipit:

Issue #67, opened 2015-12-17
gettext fails to compile in Phoenix 1.1.0 installation


cedricgc:
Hello, when trying to compile dependencies in a new Phoenix 1.1.0 application the libary gettext fails to compile. It seems to not recgonize a module called `:yecc`. I have reproduced the error with some other info that may be useful. let me know if there anything else you need.

`$ elixir --version`
Elixir 1.1.0
`$ erl -eval 'erlang:display(erlang:system_info(otp_release)), halt().'  -noshell`
"18"
`$ uname -a`
Linux worker 3.13.0-57-generic #95-Ubuntu SMP Fri Jun 19 09:28:15 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
`$ mix archive.install https://github.com/phoenixframework/phoenix/releases/download/v1.1.0/phoenix_new-1.1.0.ez`
Found existing archive(s): phoenix_new-1.0.4.ez.
Are you sure you want to replace them? [Yn] Y
creating .mix/archives/phoenix_new-1.1.0.ez
`$ mix phoenix.new example`
...
`$ cd example`
`$ cat mix.lock`
%{"connection": {:hex, :connection, "1.0.2"},
  "cowboy": {:hex, :cowboy, "1.0.4"},
  "cowlib": {:hex, :cowlib, "1.0.2"},
  "decimal": {:hex, :decimal, "1.1.0"},
  "ecto": {:hex, :ecto, "1.1.0"},
  "fs": {:hex, :fs, "0.9.2"},
  "gettext": {:hex, :gettext, "0.9.0"},
  "phoenix": {:hex, :phoenix, "1.1.0"},
  "phoenix_ecto": {:hex, :phoenix_ecto, "2.0.0"},
  "phoenix_html": {:hex, :phoenix_html, "2.3.0"},
  "phoenix_live_reload": {:hex, :phoenix_live_reload, "1.0.1"},
  "plug": {:hex, :plug, "1.0.3"},
  "poison": {:hex, :poison, "1.5.0"},
  "poolboy": {:hex, :poolboy, "1.5.1"},
  "postgrex": {:hex, :postgrex, "0.10.0"},
  "ranch": {:hex, :ranch, "1.2.0"}}
`$ mix deps.compile`
==> connection
Compiled lib/connection.ex
Generated connection app
==> poolboy (compile)
Compiled src/poolboy_worker.erl
Compiled src/poolboy_sup.erl
Compiled src/poolboy.erl
[...]

PR #72, opened 2016-01-22
Add Mix compiler that raises on missing parsetools

whatyouhide:
Not sure if this is a good idea or, if it is, if this is the way to go. I added an `:ensure_yecc_compiler` Mix compiler that raises when the `:yecc` module is missing (because the `:parsetools` Erlang application is missing). This should help a lot with issue #67 and phoenixframework/phoenix#1411.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -1,3 +1,26 @@
+defmodule Mix.Tasks.Compile.EnsureYeccCompiler do
+  def run(_args) do
+    yecc? = Code.ensure_loaded?(:yecc)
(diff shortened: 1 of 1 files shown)


lexmag:
I think we are covered with https://github.com/elixir-lang/elixir/commit/737bf1b3697f81ba3a1506c2a6802833a7a4fa00.

josevalim:
Yup, this has been fixed in Elixir. :)

whatyouhide:
Whoops, I saw that but didn't think it was this recent, I somehow just thought it didn't trigger for Gettext for some reason :\ Great then! :D

PR #73, opened 2016-02-10
Subdirectory domain support for compiler, extractor, and merger.

keichan34:


whatyouhide:
Hey @keichan34, thanks for contributing!

Do you have any evidence of this kind of domains in other Gettext implementations? I never saw support for `/`-separated domains (like `"sub/foo"` in the tests you added).

Btw, this can be accomplished with a flat domain space as well (even if it admittedly looks worse :smile:): `sub_foo.pot`, `sub_bar.pot` instead of `sub/foo.pot` and `sub/bar.pot`.

keichan34:
I wasn't able to find any other implementations that expressly support subdirectories, either.

If you think this is not appropriate for the project, I suggest at least throwing a more appropriate error when there is a path separator character in the domain, because that was causing a bad match error.

whatyouhide:
The error message absolutely has to be improved, it's terrible right now. As of the support for subdirectories, what do you think @lexmag and @josevalim? Personally, I'm slightly :-1: but the change is small so it shouldn't be too problematic in case you want to go with it.

josevalim:
I am :-1: as well since I haven't seen another impl supporting it. I think we could though check the domain for a path separator at compile time (if it is a binary at compile time) and emit a warning or an error.

josevalim:
Thank you @keichan34!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

Issue #74, opened 2016-02-10
Terrible error when POTs are inside subdirectories


whatyouhide:
When `.pot` files are inside subdirectories (e.g., `sub/foo.pot`), there's a match error which is the worst kind of error for users :).

PR #75, opened 2016-02-10
Warn when PO files are not in "canonical" path

whatyouhide:
Possible solution for #74: when compiling a `Gettext` backend and going over the PO files, we warn for all the files that are not in their "canonical" path (which is `my_locale/LC_MESSAGES/my_domain.po`).

I'm not fond of this solution as it introduces something very non-functional (emitting warnings) in an otherwise pretty functional module. Not many other solutions I can think of though. Wdyt?

### DIFF
--- a/lib/gettext/compiler.ex
+++ b/lib/gettext/compiler.ex
@@ -172,8 +172,15 @@ defmodule Gettext.Compiler do
   """
   @spec compile_po_files(Path.t) :: Macro.t
   def compile_po_files(dir) do
-    # `true` means recursively. The last argument is the initial accumulator.
-    :filelib.fold_files(String.to_char_list(dir), '\.po$', true, &compile_po_file(&1, &2), [])
+    all_po_files  = dir |> Path.join("**/*.po") |> Path.wildcard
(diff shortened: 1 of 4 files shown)


josevalim:
Here is another approach. Whenever we raise a translation not found error, check if the domain has a `/` and add a note to the error message. "We have noticed that your domain has a slash in it. Nested .po files are not supported".

whatyouhide:
That's a good (and cleaner) solution :) do we care about `.po` files outside of `LC_MESSAGES`? The solution in this PR takes care of them as well (warning that they're not being read), but if we don't care then the solution you proposed is definitely less instrusive.

josevalim:
We cannot save everyone from themselves. :) Another option is to store the domains in the backend and change the message above to check if the domain exists in the list of compiled domains and act accordingly. So we would say in the error message: "the domain xyz was not compiled. This backend is aware of the following domains: a, b, c".

whatyouhide:
I like this one, combined with checking if we have `/`-s in the domain. Will work on it tonight/tomorro!

whatyouhide:
Moved to #76!

PR #76, opened 2016-02-11
Warning for slashes in domains

whatyouhide:
Implements the ideas discussed in #75.

### DIFF
--- a/lib/gettext/compiler.ex
+++ b/lib/gettext/compiler.ex
@@ -111,14 +111,18 @@ defmodule Gettext.Compiler do
   @spec dynamic_clauses() :: Macro.t
   def dynamic_clauses do
     quote do
-      def lgettext(_, _, default, bindings) do
-        case Gettext.Interpolation.interpolate(default, bindings) do
+      def lgettext(_locale, domain, msgid, bindings) do
+        Gettext.Compiler.warn_if_domain_contains_slashes(domain)
+
+        case Gettext.Interpolation.interpolate(msgid, bindings) do
           {:ok, interpolated} -> {:default, interpolated}
           {:error, _} = error -> error
         end
       end
 
-      def lngettext(_, _, msgid, msgid_plural, n, bindings) do
+      def lngettext(_locale, domain, msgid, msgid_plural, n, bindings) do
+        Gettext.Compiler.warn_if_domain_contains_slashes(domain)
+
         str      = if n == 1, do: msgid, else: msgid_plural
         bindings = Map.put(bindings, :count, n)
 
@@ -164,6 +168,18 @@ defmodule Gettext.Compiler do
         if Enum.all?(pieces, &is_binary/1), do: Enum.join(pieces, ""), else: raiser.()
       _ ->
         raiser.()
+    end
+  end
+
+  @doc """
+  Prints a warning on `:stderr` if `domain` contains slashes.
+
+  This function is called by `lgettext` and `lngettext`.
+  """
+  @spec warn_if_domain_contains_slashes(binary) :: :ok
+  def warn_if_domain_contains_slashes(domain) do
+    if String.contains?(domain, "/") do
+      IO.puts :stderr, "warning: slashes in domains are not supported: #{inspect domain}"
     end
   end
 
--- a/lib/gettext/compiler.ex
+++ b/lib/gettext/compiler.ex
@@ -8,6 +8,7 @@ defmodule Gettext.Compiler do
 
   @plural_forms Application.get_env(:gettext, :plural_forms, Gettext.Plural)
   @default_priv "priv/gettext"
+  @po_wildcard "*/LC_MESSAGES/*.po"
 
   @doc false
   defmacro __before_compile__(env) do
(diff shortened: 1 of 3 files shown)


josevalim:
If the domain has a / in it, we can say in the error message that slashes in domain names are not supported. For this case, we may not even need to show the available domains. I don't remember why we would want to show the known domains anymore, hahaha (sorry!).

lib/gettext/compiler.ex
```diff
@@ -167,13 +174,34 @@ defmodule Gettext.Compiler do
...
+  `backend` is used to print a more helpful warning by adding a list of the
+  domains known by the backend.
+
+  This function is called by `lgettext` and `lngettext`.
+  """
+  @spec warn_if_domain_has_slashes(module, binary) :: :ok
+  def warn_if_domain_has_slashes(backend, domain) do
+    if String.contains?(domain, "/") do
+      IO.puts :stderr, """
+      The domain #{inspect domain} was not compiled. The #{inspect backend}
```
josevalim: warning: the domain ...

whatyouhide:
@josevalim makes sense :) I squashed the second commit, now we warn when the domain is unknown (and we show the available domains) and we warn that `/`s in domains aren't supported if the domain contains any `/`s.

lib/gettext/compiler.ex
```diff
@@ -167,13 +174,34 @@ defmodule Gettext.Compiler do
...
+
+  This function is called by `lgettext` and `lngettext`.
+  """
+  @spec warn_unknown_domain(module, binary) :: :ok
+  def warn_unknown_domain(backend, domain) do
+    known = backend.__gettext__(:known_domains)
+    cond do
+      String.contains?(domain, "/") ->
+        IO.puts :stderr, "warning: slashes in domains are not supported: #{inspect domain}"
+      not domain in known ->
```
josevalim: Discussion point: should we really warn here? Should we ask folks to create empty .po files in their locales then? I would say we shouldn't warn here (we can still keep known_domains around, I would just remove the warning. If we warn, we will break the workflow where I can just add translations, specially in english, without caring about the .po / .pot files.

josevalim:
Sorry I was not super clear in my previous comment(s).  My point is simply that we can keep known_domains around but I don't think we would use it for anything right now. I think the warning will cause more confusion than help. :)

josevalim:
And to be super clear: the warning on slashes is still super useful.

lib/gettext/compiler.ex
lexmag: I'm :+1: for not having this warning or make it configurable, e.g. `allow_missing: false`. :bowtie:

lib/gettext/compiler.ex
```diff
@@ -167,13 +174,34 @@ defmodule Gettext.Compiler do
   end
 
   @doc """
+  Prints a warning on `:stderr` if `domain` isn't known to `backend`.
+
+  `backend` is used to print a more helpful warning by adding a list of the
+  domains known by the backend.
+
+  This function is called by `lgettext` and `lngettext`.
+  """
```
lexmag: Maybe it should be `@doc false`?

whatyouhide: The whole module is `@moduledoc false`, all other functions are documented with `@doc`.

lib/gettext/compiler.ex
whatyouhide: Honestly, I don't want to warn on missing domains. It's a little bit of a mess, we have to capture so much IO in the tests and @josevalim makes a great point (you'll be warned **everytime** until you run `mix gettext.extract`).

whatyouhide:
Ok, we're superclear now: we'll keep the known locales around (cause it's easy and it can be useful) and we'll only warn for domains with `/`s in them.

lexmag:
> ... we'll keep the known locales around (cause it's easy and it can be useful) ...

You mean known domains, right?
If so, why we keep them if there is no use for it – dead code.

josevalim:
Feel free to remove known_domains. but it doesn't hurt to keep it as well.

whatyouhide:
Ok @josevalim and @lexmag, I decided to remove `known_domains`-related stuff as it's indeed dead code and it's something internal to Gettext, so it wouldn't benefit users; we can always re-add it later (as it's terribly simple to re-implement).

I also fixed the warnings: now only domains with slashes in them emit warnings when used. Ready to merge?

josevalim:
:shipit:

vborodkin:
Hi there! I might be missing something, but how do you treat nested domains then? For example, `dgettext "users/admin", "List Users"`, where the domain file should inside a `users` directory.

whatyouhide:
@vborodkin is this something GNU gettext supports?

vborodkin:
@whatyouhide I don't know. I never used gettext until I started using Phoenix, which is recently. Is there anything I can do to help and clarify this? Although, it seems fairly normal to namespace domains in larger applications. I can use a underscore for the time being.

OvermindDL1:
GNU gettext does support `/` in domains just for note (I just tested it).

PR #78, opened 2016-02-16
Better comment handling

whatyouhide:
We're now:

* adding headers to new POT files (we already added them to new PO files), so that informative comments are kept as the result of being attached to the headers
* handling "dangling" comments, i.e., comments that are not followed by translations (we're simply throwing them away)

### DIFF
--- a/lib/gettext/extractor.ex
+++ b/lib/gettext/extractor.ex
@@ -176,10 +176,11 @@ defmodule Gettext.Extractor do
   defp tag_files({path, new_po}),
     do: {path, {:new, new_po}}
 
-  # This function dumps merged files and unmerged files without any changes, and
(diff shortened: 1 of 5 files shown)


lib/gettext/extractor.ex
```diff
@@ -144,12 +144,12 @@ defmodule Gettext.Extractor do
 
   # Returns :unchanged if merging `existing_path` with `new` changes nothing,
   # otherwise a %Gettext.PO{} struct with the changed contents.
-  defp merge_or_unchanged(existing_path, new) do
+  defp merge_or_unchanged(existing_path, new_struct) do
```
lexmag: :clap: for better name.

josevalim:
:shipit:

PR #81, opened 2016-02-25
Plural enhancements

whatyouhide:
@josevalim I kept `pt_BR` to be compliant with [this](https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_and_Plurals) and [this](http://localization-guide.readthedocs.org/en/latest/l10n/pluralforms.html), and because the implementation isn't really any more complicated IMO. I can remove it and treat `pt_BR` as `pt` if you prefer. Wdyt?

### DIFF
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -349,11 +349,13 @@ defmodule Gettext do
 
   ## Configuration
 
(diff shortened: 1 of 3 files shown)


josevalim:

PR #82, opened 2016-02-26
Better plural forms

whatyouhide:
We were reading the `:plural_forms` option (that specifies which pluralizer module to use) from the configuration of the `:gettext` application, but we want to read that from the configuration of each backend (the one passed to `use Gettext`).

I fixed that and also polished the modules for `Gettext.Plural`.


josevalim:
:shipit:

PR #84, opened 2016-03-17
Add the Gettext.Backend behaviour

whatyouhide:
This behaviour is implemented by all modules that call `use Gettext`. It's mostly useful for documentation. I haven't included the `lgettext/4` and `lngettext/6` callbacks in the behaviour as we keep those function "hidden" (not documented anywhere currently); however, we call those from the `Gettext.*gettext` functions so it may be worth to add them to the behaviour, so that if people want to implement weird Gettext backends they have *all* the correct callbacks.

Wdyt @lexmag and @josevalim?

### DIFF
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -115,7 +115,9 @@ defmodule Gettext do
 
   ### Using macros
 
-  When a module calls `use Gettext`, the following macros are automatically
+  Each module that calls `use Gettext` is usually referred to as a "Gettext
+  backend", as it implements the `Gettext.Backend` behaviour. When a module
+  calls `use Gettext`, the following macros are automatically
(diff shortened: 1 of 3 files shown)


lexmag:
Nice. :yellow_heart:

josevalim:
I would define them in Gettext, which is the module that we use. WDYT?

whatyouhide:
@josevalim I started out in `Gettext`, but realized `Gettext` already has so much documentation that I was afraid these callbacks wouldn't stand out enough :). That's my only point in favour of `Gettext.Backend`; I can move the behaviour to Gettext of course. Btw, wdty about the `l(n)gettext` stuff?

josevalim:
:shipit:

Issue #85, opened 2016-03-30
Msgids are deleted from POT files


maufl:
I read through the issues and found that it's a feature, but unfortunately not for me.
When running `mix gettext.extract` all msgids from POT files that have a reference are deleted if they are not present in the extracted msgids.

I'm trying to extract gettext msgids from React components that we use in our Phoenix project.
Phoenix uses Webpack to bundle frontend assets and I found this handy [loader](https://github.com/nnarhinen/jsxgettext-loader) that will extract msgids from Js(x) files when they are loaded via webpack.
All the msgids are dumped to a new POT file in priv/gettext (because I configured it so) and they all have references.
Now when I extract msgid using mix, the POT file is emptied :(

I can't change the behaviour of the loader without patching jsxgettext and jsxgettext-loader.
I could put the POT file in another directory, but right now the workflow for Phoenix and Reactis perfectly in sync and I would like to keep it that way.

Is there a possibility to force (Elixir) gettext to not delete msgids from a POT file?

josevalim:
@maufl can you please paste some examples of the msgids extracted from jsxgettext? Maybe we can provide some directory whitelisting. This way we won't remove msgids if they belong to a certain reference.

maufl:
Currently it looks like this

```
msgid ""
msgstr ""
"Project-Id-Version: PACKAGE VERSION\n"
"Language-Team: LANGUAGE <LL@li.org>\n"
"Report-Msgid-Bugs-To: \n"
"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
"Language: \n"
"MIME-Version: 1.0\n"
"Content-Type: text/plain; charset=utf-8\n"
"Content-Transfer-Encoding: 8bit\n"
"POT-Creation-Date: 2016-03-30 12:35+0000\n"

#: 0:91
msgid "Column 1"
msgstr ""

#: 0:96
msgid "Column 2"
msgid_plural "Columns 2"
msgstr[0] ""
msgstr[1] ""
```

But I opend a [pull request](https://github.com/nnarhinen/jsxgettext-loader/pull/1) in the loader repo that should fix the file names, e.g. instead of `0` it will be `web/static/js/react_test.js`.

Directory whitelisting sounds like a good idea to me.

maufl:
The change landed in the loader and now the file looks like this:

```
msgid ""
msgstr ""
"Project-Id-Version: PACKAGE VERSION\n"
[...]

PR #86, opened 2016-04-04
Allow whitelisting of references

maufl:
Adds an option to specify a pattern of references that should be preserved by Gettext.
As discussed in #85.


lib/gettext/po/translations.ex
```diff
@@ -34,6 +34,31 @@ defmodule Gettext.PO.Translations do
...
   @doc """
+  Tells whether a translation is protected from purging.
+
+  A translation that is protected from purging will never be removed by Gettext.
+  Which translations are proteced can be configured using Mix.
+
+  ## Example
+
+      config :gettext,
+        exlude_refs_from_pruging: "^web\/static\/.*"
```
whatyouhide: Typo in `_pruging`.

lib/gettext/po/translations.ex
```diff
@@ -34,6 +34,31 @@ defmodule Gettext.PO.Translations do
...
+      
+      t = %Gettext.PO.Translation{msgid: "Hello world!", references: [{"web/static/js/app.js", 42}]}
+      Gettext.PO.Translations.protected?(t) == true
+  """
+  @spec protected?(Gettext.PO.translation) :: boolean
+  def protected?(%{__struct__: s, msgid: msgid, references: []}) when is_translation(s), do: false
+
+  def protected?(%{__struct__: s, msgid: msgid, references: refs}) when is_translation(s) do
+    {:ok, pattern} = Application.get_env(:gettext, :excluded_refs_from_purging, "(?!x)x")
+    |> Regex.compile
```
whatyouhide: Why are we doing this string regex + `Regex.compile/1` dance? Can't we just use `~r//` in the configuration files?

lib/gettext/po/translations.ex
```diff
@@ -34,6 +34,31 @@ defmodule Gettext.PO.Translations do
...
+  Tells whether a translation is protected from purging.
+
+  A translation that is protected from purging will never be removed by Gettext.
+  Which translations are proteced can be configured using Mix.
+
+  ## Example
+
+      config :gettext,
+        exlude_refs_from_pruging: "^web\/static\/.*"
+
```
whatyouhide: Trailing whitespace.

lib/gettext/po/translations.ex
```diff
@@ -34,6 +34,31 @@ defmodule Gettext.PO.Translations do
...
+      Gettext.PO.Translations.protected?(t) == true
+  """
+  @spec protected?(Gettext.PO.translation) :: boolean
+  def protected?(%{__struct__: s, msgid: msgid, references: []}) when is_translation(s), do: false
+
+  def protected?(%{__struct__: s, msgid: msgid, references: refs}) when is_translation(s) do
+    {:ok, pattern} = Application.get_env(:gettext, :excluded_refs_from_purging, "(?!x)x")
+    |> Regex.compile
+    refs
+    |> Enum.map(fn({line, _}) -> line end)
```
whatyouhide: This should be called `path` (or `file`) instead of `line`?

lib/gettext.ex
```diff
@@ -366,6 +366,12 @@ defmodule Gettext do
       gettext.merge` task (`Mix.Tasks.Gettext.Merge`) for more information on
       fuzzy translations.
 
+    * `:excluded_refs_from_purging` - a regex pattern as a string. Gettext will
+      ignore and preserve all translations in all .pot files that have a reference
+      that matches this pattern. You can use this pattern to prevent Gettext from
+      removing translations that you have extracted using another tool, e.g. from
+      static resources (web/static).
```
whatyouhide: Let's not mention `web/static`, we can just say

> [...] that you have extracted using another tool.

test/gettext/extractor_test.exs
```diff
@@ -62,6 +62,17 @@ defmodule Gettext.ExtractorTest do
     assert Extractor.merge_template(old, new) == %PO{translations: [t1, t2, t3]}
   end
 
+  test "merge_template/2: whitelisted translations are kept" do
+    # Exclude translations from static resources
+    Application.put_env(:gettext, :excluded_refs_from_purging, "^web\/static\/.*$")
```
whatyouhide: This regex can be simplified to be `^web/static/`.

lib/gettext/po/translations.ex
whatyouhide: Also, not sure why you're escaping `/` given you're in a string; it would make sense in `~r//`, but in strings you can just use `^web/static/.*`. :)

whatyouhide:
Hey @maufl, great job, this is definitely headed in the right direction. I left a first batch of comments, let me know if everything is clear :(

lib/gettext/po/translations.ex
maufl: I don't know why exactly I did it but I think it was because I first implemented an option to ignore a domain. I added this option to the options of a backend and found that I can not unquote a regex, so I passed a string.
I agree that a regex in the configuration is nicer, I made a change that fixes that.

maufl:
Argh, this damn pruging typo followed me everywhere, I thought I found all of them ...

I made some improvements, I think they cover all your suggestions.

lib/gettext.ex
```diff
@@ -366,11 +366,10 @@ defmodule Gettext do
       gettext.merge` task (`Mix.Tasks.Gettext.Merge`) for more information on
       fuzzy translations.
 
-    * `:excluded_refs_from_purging` - a regex pattern as a string. Gettext will
-      ignore and preserve all translations in all .pot files that have a reference
-      that matches this pattern. You can use this pattern to prevent Gettext from
-      removing translations that you have extracted using another tool, e.g. from
-      static resources (web/static).
+    * `:excluded_refs_from_purging` - a regex that is matched agains translation
```
whatyouhide: `agains` -> `agains`

lib/gettext.ex
```diff
@@ -366,11 +366,10 @@ defmodule Gettext do
...
+      references. Gettext will preserve all translations in all .pot files that
```
whatyouhide: Can we use POT instead of .pot? Or `.pot` (with backticks around) is fine as well.

lib/gettext/po/translations.ex
```diff
@@ -42,8 +42,8 @@ defmodule Gettext.PO.Translations do
   ## Example
 
       config :gettext,
-        exlude_refs_from_pruging: "^web\/static\/.*"
-      
+        exclude_refs_from_purging: ~r/^web\/static\/.*/
```
whatyouhide: You don't have to use `/` as a delimiter for the `~r` sigil; here, you could use `{}` so that you don't have to escape `/`: `~r{^web/static}`.

lib/gettext/po/translations.ex
```diff
@@ -51,10 +51,9 @@ defmodule Gettext.PO.Translations do
   def protected?(%{__struct__: s, msgid: msgid, references: []}) when is_translation(s), do: false
 
   def protected?(%{__struct__: s, msgid: msgid, references: refs}) when is_translation(s) do
-    {:ok, pattern} = Application.get_env(:gettext, :excluded_refs_from_purging, "(?!x)x")
-    |> Regex.compile
+    pattern = Application.get_env(:gettext, :excluded_refs_from_purging, ~r/(?!x)x/)
     refs
-    |> Enum.map(fn({line, _}) -> line end)
+    |> Enum.map(fn({path, _}) -> path end)
     |> Enum.any?(&Regex.match?(pattern, &1))
```
whatyouhide: We can squeeze `Enum.map/2` + `Enum.any?/2` into a single pass of `Enum.any?/2`:

```elixir
Enum.any?(refs, fn({path, _}) -> Regex.match?(pattern, path) end)
```

test/gettext/extractor_test.exs
```diff
@@ -64,7 +64,7 @@ defmodule Gettext.ExtractorTest do
 
   test "merge_template/2: whitelisted translations are kept" do
     # Exclude translations from static resources
-    Application.put_env(:gettext, :excluded_refs_from_purging, "^web\/static\/.*$")
+    Application.put_env(:gettext, :excluded_refs_from_purging, ~r/^web\/static\//)
```
whatyouhide: Same as above regarding escaping.

maufl:
I love how all you suggestions make the code smaller and simpler.

lib/gettext/po/translations.ex
```diff
@@ -42,7 +42,7 @@ defmodule Gettext.PO.Translations do
   ## Example
 
       config :gettext,
-        exclude_refs_from_purging: ~r/^web\/static\/.*/
+        exclude_refs_from_purging: ~r{^web/static/.*}
```
whatyouhide: Maybe better to make this consistent with the rest of the regexes, `~r{^web/static/}`.

whatyouhide:
@maufl awesome. So, I've been thinking about this and I think we should move the impure logic of this up from where it is now. By impure logic, I mean the retrieving of the `:exclude_refs_from_purging` option from `:gettext`'s environment. We should move that up at least to `Gettext.Extractor.merge_template/2`, e.g.:

```elixir
def merge_template(existing, new) do
  protected_pattern = Application.get_env(:gettext, :exclude_refs_from_purging)
  ...
    PO.Translations.protected?(t, protected_pattern)
```

This way, `protected?/2` becomes simpler to reason about and you only retrieve the pattern from the app config once (instead of once for each translation).

Also, I don't love the idea of having a "fallback" regex that never matches when the `:exclude_refs_from_purging` option is not provided; personally, I'd like more something along these lines:

```elixir
protected_pattern = Application.get_env(:gettext, :exclude_refs_from_purging)

# in protected?/2
def protected?(_t, nil), do: false
def protected?(_t, %Regex{} = pattern), do: ...
```

Wdyt?

maufl:
I agree that it's ugly to fetch the pattern every time the `protected?/1` method is called, but I'm not really sure if I like it better in `merge_template/2`.

Originally I used the fallback regex to avoid conditional expressions in `protected?/1`. Your proposal avoids conditional logic in in the function anyway.  I think it's a better solution than what I did so far.

whatyouhide:
@maufl why you don't like the idea of fetching the option in `merge_template/2`? Or better, why do you like it better in `protected?/1`?

(btw, ideally we would have this as up as possible, e.g., in the `gettext.extract` Mix task, that later passes this option around so that also `merge_template/2` can take this option and become pure, but let's do small steps here)

maufl:
I don't like it better in `protected?/1`, but I'm also not sure where I would put it :/ I'm lacking Elixir experience to have a good feeling about where it should go.

whatyouhide:
@maufl let's put it in `merge_template/2` and pass it along to `protected?/2` for now, we can always revise later.

We're almost good to go :D We're only missing docs for this new option. You can describe this option in [this section](https://github.com/elixir-lang/gettext/blob/master/lib/gettext.ex#L356-L367) of the docs.

maufl:
I already added docs for this option. Should be the first changeset when you compare the files. :)

whatyouhide:
Ooops, very sorry :) I even left a few comments there :smile: So, very very last thing: can you squash all the commits in a single one? Then I think we're ready to go :+1:

maufl:
You mean, squash them locally and force push? I just discovered in another repo that GitHub allows you to squash commits of a pull request before merging them. If you hit "Merge pull request" you will get a "Confirm merge" button with a dropdown. The squash options is hidden in the dropdown.

lib/gettext/po/translations.ex
```diff
@@ -34,6 +34,25 @@ defmodule Gettext.PO.Translations do
...
+
+  A translation that is protected from purging will never be removed by Gettext.
+  Which translations are proteced can be configured using Mix.
+
+  ## Example
+
+      iex> protected_pattern = ~r{^web/static/}
+      iex> t = %Gettext.PO.Translation{msgid: "Hello world!", references: [{"web/static/js/app.js", 42}]}
+      iex> Gettext.PO.Translations.protected?(t, protected_pattern)
+      true
```
whatyouhide: Can we leave an empty line after the `true`? It's very common to do that in Elixir.

lib/gettext/po/translations.ex
```diff
@@ -34,6 +34,25 @@ defmodule Gettext.PO.Translations do
...
+  """
+  @spec protected?(Gettext.PO.translation, Regex.t) :: boolean
+  def protected?(_t, nil), do: false
+  def protected?(%{__struct__: s, references: []}, pattern) when is_translation(s), do: false
+  def protected?(%{__struct__: s, references: refs}, pattern) when is_translation(s),
+    do: Enum.any?(refs, fn({path, _}) -> Regex.match?(pattern, path) end)
```
whatyouhide: This is great with `do:` instead of `do ... end` :) Last nitpick: can we always have the `do:` on a new line?

```elixir
  def protected?(_t, nil),
    do: false
  def protected?(%{__struct__: s, references: []}, pattern) when is_translation(s),
    do: false
  def protected?(%{__struct__: s, references: refs}, pattern) when is_translation(s),
    do: Enum.any?(refs, fn({path, _}) -> Regex.match?(pattern, path) end)
```

whatyouhide:
oh, you're right, force of habit :) Left the superlast couple of comments, then we're ready to merge :)

lib/gettext/po/translations.ex
maufl: Of course we can :)

maufl:
Fixed :) Somehow Emacs Elixir mode has issues with this kind of indentation :/

whatyouhide:
Wonderful :heart_decoration: Thanks so much!

(btw yep, this is hard code to indent I think, that's why elixir-mode has trouble with it)

maufl:
Your welcome :) Feels great already to contribute to an Elixir library after only about 2 months.

Issue #87, opened 2016-04-05
Default pluralizer does not have fallback for unknown locales


maufl:
Hi, it's me again ^^
The documentation says:

> Locale
> 
> At runtime, all gettext-related functions and macros that do not explicitely take a locale as an argument read the locale from Gettext.get_locale/1. The locale can be set with Gettext.put_locale/2. Locales are expressed as strings (like "en" or "fr"); they can be arbitrary strings as long as they match a directory name.

However, I just found that this locale will be passed to `Gettext.Plural.plural` if no other pluralizer is configured. I get a `no function clause matching` for the locale `nb-no`.

I think there should be a catch all clause to avoid problems when people use strange locales.

whatyouhide:
Ok, note to self (so that I understand this better :smiley:):

1. A user wants to use a locale that `Gettext.Plural` doesn't cover, e.g., `nb-no`.
2. They add the `nb-no/LC_MESSAGES/*.po` translation files, hence...
3. ...a `nb-no` clause is generated for the `l(n)gettext` functions in each backend.
4. This clause is hit when translating `nb-no` messages, so we hit [these lines](https://github.com/elixir-lang/gettext/blob/784d97eb6d897a89843c896c8d5d3cb1be9081f4/lib/gettext/compiler.ex#L250-L256), where `@plural_forms.plural("nb-no", n)` is called.
5. `Gettext.Plural` doesn't cover `"nb-no"`, thus function clause error :)

@josevalim thoughts on what to do here? We could have a match-all clause in `Gettext.Plural` and warn, returning English pluralization rules in the meantime, or a match-all clause that simply raises. Wdyt?

josevalim:
It is probably best to raise with a good error message and instructions on how to install your own plural rules. Right?

whatyouhide:
@josevalim right, I even updated the docs for `Gettext.Plural` some time ago exactly to show how to just add a locale and fallback to `Gettext.Plural` for all others. Will do this in a moment then.

whatyouhide:
@josevalim a small problem arised while implementing this: right now, since we do `locale <> _` for all clauses, I had trouble reproducing this because

```elixir
Gettext.Plural.plural("wat", 1)
```

[...]

PR #88, opened 2016-04-05
Handle unknown locales in Gettext.Plural

whatyouhide:
Implements what was discussed in #87. @josevalim let me know what you think, but it should be straightforward :)

/cc @maufl

### DIFF
--- a/lib/gettext/plural.ex
+++ b/lib/gettext/plural.ex
@@ -62,6 +62,22 @@ defmodule Gettext.Plural do
   as well as from [Mozilla's guide on "Localization and
   plurals"](https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_and_Plurals).
 
+  ### Unknown locales
+
+  Trying to call `Gettext.Plural` functions with unknown locales will result in
+  a `UnknownLocaleError` exception.
+
+  ### Language and country
+
+  Often, a locale is composed as a language and country couple, such as
+  `en_US`. The default implementation for `Gettext.Plural` handles `xx_YY` by
+  forwarding it to `xx` (except for *just Brazilian Portuguese*, `pt_BR`, which
+  is not forwarded to `pt` as pluralization rules slightly differ). We treat the
+  underscore as a separator according to ISO 15897. Sometimes, a dash `-` is
+  used as a separator (e.g., `en-US`): this is not forwarded to `en` in the
+  default `Gettext.Plural` (and it will raise an `Gettext.Plural.UnknownLocaleError`
+  exception).
+
   ## Examples
 
   An example of the plural form of a given number of elements in the Polish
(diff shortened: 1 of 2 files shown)


maufl:
Looks good to me, except that it will not handle locales separated by a dash right?

whatyouhide:
@maufl nope, as I'm not sure that's a common use case. You can fix this by hand extremely easily anyways by creating a new pluralizer module.

maufl:
Are you sure? Because it seems that this is the format used in accept language headers.

But of course it would work for me too :)

josevalim:
@maufl in this case it likely makes sense to convert to underscore before passing to gettext. :)

whatyouhide:
@maufl has a point though:

* underscores are fine: https://en.wikipedia.org/wiki/Locale
* dashes too: https://en.wikipedia.org/wiki/IETF_language_tag

not sure we should support both. I'd of course prefer to support just underscores as the implementation is cleaner, stricter and less error-prone, but I'm not convinced.

josevalim:
Yes, I'd go with one format anyway otherwise you end-up having inconsistent locale folders between projects. Imagine having to set "pt_BR" in one backend and "pt-BR" in another locale backend.

whatyouhide:
Let's go with only underscores then, I'll updated the docs and then merge this.

whatyouhide:
@josevalim updated the PR with docs and a new exception, `Gettext.Plural.UnknownLocaleError`, that can be raised when a pluralized module doesn't know how to handle a locale:

```elixir
raise Gettext.Plural.UnknownLocaleError, "no-nb"
```

would love a quick review before merging (of course, @lexmag can pitch in if he wants :smiley:).

@maufl by the way, implementing a custom pluralizer module that handles dashes should be straightforward and a very elegant solution IMO :)

```elixir
defmodule DashPluralized do
  @behaviour Gettext.Plural

  def nplurals(locale) do
    locale
    |> String.split("-", parts: 2, trim: true)
    |> hd()
    |> Gettext.Plural.nplurals()
  end

  def plural(locale, n) do
    # pretty much the same thing
  end
end
```

josevalim:
It looks great. I still wouldn't suggest a custom pluralizer for handling the dashes because it would still mean you would need to have no_OT and no-OT in the filesystem, no?

whatyouhide:
@josevalim oh, good point. I keep forgetting that pluralizing and translating are two different things :) but for @maufl's use case, if he only uses `no-OT`, then `DashPluralizer` would still be a good fit IMO.

I updated the docs so that they don't mention anything about implementing a forwarding pluralizer :)

maufl:
I don't understand why I would need to have folders with dashes and underscores, but thank you very much for the code snippet @whatyouhide :) I'll add that to our code base next week.

whatyouhide:
@maufl if you want to support both `no_OT` and `no-OT`, you'd have to have translations in both the `no_OT` as well as the `no-OT

maufl:
Ok, I did not even assume that anybody would do something like that, without normalizing it in the application itself ...

Issue #89, opened 2016-04-08
Syntax errors can be absolutely terrible


whatyouhide:
A PO file like this beauty (`default.po`):

``` po
#: lib/foo.ex:1
msgid "hello"

#: lib/foo.ex:2
msgid "world"
msgstr ""
```

Spits out a basically impossible-to-decipher error:

```
** (Gettext.PO.SyntaxError) 4: syntax error before: <<"#: lib/foo.ex:2">>
```

Why this sucks (as if it was need to show explicitly :stuck_out_tongue:):
- [x] no file name (wat!!!)
- [x] Erlang binary instead of Elixir binary
- [ ] finally, it should just say something like "missing `msgstr` for `msgid "hello"`"

@josevalim I'm afraid the only way to do this is to move the logic that builds translations from msgids and msgstrs outside of the `.yrl` parser, in Elixir-land? Wdyt?

josevalim:
@whatyouhide it can be fixed, check how Elixir does it, but it is not pretty. :) In particular, if you can stick to erlang like tokens, things will be better.

josevalim:
Actually, didn't you work on a PR in Elixir's parser to solve similar issues? :)

whatyouhide:
@josevalim indeed I did this for char literals a few weeks ago in Elixir, but the Erlang representation here is just one of the problems. We can fix it by rewriting errors that match, e.g., `<<` but other than it not being pretty, it doesn't solve what I believe is a more important issue, which is that maybe we should probably just say "missing `msgstr` for `msgid "hello"`" as an error message instead of a plain syntax error. Wdyt?

josevalim:
One option is to simplify the parsing to only parse lines, so it would
return a list of ["comment", "msgstr", "msgid"], and you would do the
grouping and post-validation later in a separate step.



*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

whatyouhide:
@josevalim sorry, I wasn't very clear earlier, but what you suggested is exactly what I meant with

>  I'm afraid the only way to do this is to move the logic that builds translations from msgids and msgstrs outside of the `.yrl` parser, in Elixir-land

:) I'll see what I can come up with. :smiley:

tmbb:
NimbleParsec is oretty sweet for writing `.po` files amd by deafult it returns some pretty readable errors. I assume errors can be made even better with some careful grammar design.

[...]

PR #90, opened 2016-04-08
Show the file name in syntax errors when extracting

whatyouhide:
The current `Gettext.Extractor` code manually reads files through `File.read!/1` and then parses them with `Gettext.PO.parse_string!/1`. The problem with this is that if there's a syntax error, then the file name where the error is won't appear in the error message as the error is raised by `parse_string!/1`.

This commit fixes this by still calling `File.read!/1`, but also calling `parse_file!/1` after that (so effectively reading the file twice, which should be fine as we don't care for raw speed here).

@josevalim what do you think about this to solve one of the problems listed in #89? Again, yeah we read the file twice but the alternative is something like this:

```elixir
    existing_po =
      case PO.parse_string(existing_contents) do
        {:ok, po} ->
          po
        {:error, line, reason} ->
          raise PO.SyntaxError, file: existing_path, line: line, reason: reason
      end
    merged = merge_template(existing_po, new_struct)
```

where we manually raise `Gettext.PO.SyntaxError` with file, line and reason. Not sure about this, wdyt?

(btw, I'd love @lexmag's opinion on this as well as he's the one who found the bug today :smiley:)


josevalim:
I would go with the example where we read it once because that's exactly what we do in Elixir. :D

lexmag:
Reading file once seems better. :bowtie:

whatyouhide:
Updated with the manual-raising code, looks easy to read indeed so yay :smiley:

PR #92, opened 2016-04-22
Allow duplicate interpolation keys

chazsconi:
This fixes a bug where it is not possible to specify the same interpolation key twice in a string
e.g. `"Hello %{name}. Goodbye %{name}"`

### DIFF
--- a/lib/gettext/interpolation.ex
+++ b/lib/gettext/interpolation.ex
@@ -70,13 +70,16 @@ defmodule Gettext.Interpolation do
       iex> Gettext.Interpolation.keys(["Hello ", :name, "!"])
       [:name]
 
+      iex> Gettext.Interpolation.keys(["Hello ", :name, "! Goodbye", :name])
+      [:name]
+
   """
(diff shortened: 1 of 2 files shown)


lib/gettext/interpolation.ex
```diff
@@ -70,13 +70,16 @@ defmodule Gettext.Interpolation do
       iex> Gettext.Interpolation.keys(["Hello ", :name, "!"])
       [:name]
 
+      iex> Gettext.Interpolation.keys(["Hello ", :name, "!", " goodbye", :name])
```
whatyouhide: Just a nitpick, but can we put `"!"` and `" goodbye"` together? Something like:

```elixir
["Hello ", :name, "! Goodbye, ", :name]
```

whatyouhide:
Great catch @chazsconi, thanks for the contribution 😃 💟 I just left one small nitpicky comment and we're ready to merge :)

chazsconi:
Sure - fixed.

whatyouhide:
💟

PR #97, opened 2016-05-13
Raise when the domain is not a binary at compile-time

### LINKED ISSUE #94: Extraction fails when using variable domain
As per https://groups.google.com/forum/#!topic/elixir-lang-talk/GyYMj_4oc4o, `mix gettext.extract` crashes when used with a variable domain name, such as `dgettext(domain, "Say something")`

```
** (Protocol.UndefinedError) protocol String.Chars not implemented for {:domain, [line: 7], nil}
    (elixir) lib/string/chars.ex:3: String.Chars.impl_for!/1
    (elixir) lib/string/chars.ex:17: String.Chars.to_string/1
    lib/gettext/extractor.ex:93: Gettext.Extractor.pot_path/2
    lib/gettext/extractor.ex:89: Gettext.Extractor.create_po_struct/3
    lib/gettext/extractor.ex:83: anonymous fn/3 in Gettext.Extractor.create_po_structs_from_extracted_translations/1
    (stdlib) lists.erl:1262: :lists.foldl/3
    lib/gettext/extractor.ex:81: Gettext.Extractor.create_po_structs_from_extracted_translations/1
    lib/gettext/extractor.ex:62: Gettext.Extractor.pot_files/0
    lib/mix/tasks/gettext.extract.ex:59: Mix.Tasks.Gettext.Extract.extract/0
    lib/mix/tasks/gettext.extract.ex:30: Mix.Tasks.Gettext.Extract.run/1
    (mix) lib/mix/cli.ex:58: Mix.CLI.run_task/2
```
Preferably I would not have it crash but rather just skip over the calls it cannot extract so that I can create those myself, and fill in the calls that it did extract.

whatyouhide:
Should close #94.

### DIFF
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -156,8 +156,8 @@ defmodule Gettext do
       #=> "Ciao mondo"
 
   The `gettext`/`dgettext`/`ngettext`/`dngettext` macros raise an
-  `ArgumentError` exception if they receive a `msgid` or a `msgid_plural` that
-  doesn't expand to a string at compile time:
+  `ArgumentError` exception if they receive a `domain`, `msgid`, or
+  `msgid_plural` that doesn't expand to a string at compile time:
 
       msgid = "Hello world"
       MyApp.Gettext.gettext msgid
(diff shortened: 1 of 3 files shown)


josevalim:
:shipit: :heart:

PR #100, opened 2016-05-19
Add a :compiler_po_wildcard option to the Gettext configruation

### LINKED ISSUE #99: Using wildcard in the gettext compiler can be really slow
Brought up by https://github.com/phoenixframework/phoenix/issues/1671.

whatyouhide:
This should fix #99.

Right now, the `:gettext` compiler uses a hardcoded default wildcard of `"**/*.po"` to find all the PO files under `priv` (so that manifests can be built and updated). This turns out to be very slow when there are a lot of files in `priv`, as it may happen for example in a Phoenix application with a lot of stuff in `priv/vendor` (see https://github.com/phoenixframework/phoenix/issues/1671).

This commit adds support for the `:compiler_po_wildcard` option to the `:gettext` application. This option specifies the wildcard that the compiler uses  to find PO files (relative to `priv`). It defaults to `gettext/*/LC_MESSAGES/*.po`, which is in line with the default directory for Gettext backends, which is `priv/gettext`.

I'm not 100% sure about the name of this option, but I think it's a start. Wdyt @josevalim?

(on an unrelated note, PR number 100! 🎉)

### DIFF
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -375,6 +375,12 @@ defmodule Gettext do
       have a matching reference. You can use this pattern to prevent Gettext from
       removing translations that you have extracted using another tool.
 
+    * `:compiler_po_wildcard` - a binary that specifies the wildcard that the
+      `:gettext` compiler will use to find changed PO files in order to recompile
+      their respective Gettext backends. This wildcard has to be relative to the
+      `"priv"` directory of your application. Defaults to
+      `"gettext/*/LC_MESSAGES/*.po"`.
+
   ### Backend configuration
 
   A Gettext backend such as `MyApp.Gettext` can be configured under the
--- a/lib/mix/tasks/compile.gettext.ex
+++ b/lib/mix/tasks/compile.gettext.ex
@@ -15,13 +15,18 @@ defmodule Mix.Tasks.Compile.Gettext do
   # that each backend specifies the corresponding manifest file as an
   # @external_resource.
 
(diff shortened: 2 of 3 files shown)


josevalim:
I am not sure about "gettext/*/LC_MESSAGES/*.po" because in theory gettext is tied to the backend and if you have multiple backends they will exist at different namespaces.

One option is to try to find all gettext backends in the codepath and only look there but that's tricky. Another option is to log somewhere every time we compile a gettext backend and use that information to know where to look at.

> (on an unrelated note, PR number 100! 🎉)

🎉

whatyouhide:
Finding all backends in the codepath sounds painful, definitely. My idea with `gettext/*/LC_MESSAGES/*.po` was that this is the default location for `:priv` for Gettext backends, so if you just do everything according to defaults then everything will play nice;  if you customize the `:priv` for a backend, you'll have to update your `:compiler_po_wildcard` config as well.

However, writing down the Gettext backends may work but I don't think we can do that on the first run (since we suggest to put the `:gettext` compiler before other Mix compilers). Wdyt?

josevalim:
There is also the issue we are configuring the gettext application which means that if two different projects use two different lookup patterns, it won't work. :( Do we store any information in the user application today?

whatyouhide:
@josevalim updated this to use the Mix configuration. Let me know what you think :)

josevalim:
Beautiful, ship it!

PR #102, opened 2016-05-28
Don't dig into the directory given to gettext.merge

whatyouhide:
\cc @josevalim, point 1. of what we discussed a while back :)


josevalim:

PR #105, opened 2016-06-21
Add Tagalog(tl) pluralization rules

Rakoth:


mix.lock
```diff
@@ -1,2 +1,2 @@
-%{"earmark": {:hex, :earmark, "0.2.1"},
-  "ex_doc": {:hex, :ex_doc, "0.11.4"}}
+%{"earmark": {:hex, :earmark, "0.2.1", "ba6d26ceb16106d069b289df66751734802777a3cbb6787026dd800ffeb850f3", [:mix], []},
+  "ex_doc": {:hex, :ex_doc, "0.11.4", "a064bdb720594c3745b94709b17ffb834fd858b4e0c1f48f37c0d92700759e02", [:mix], [{:earmark, "~> 0.1.17 or ~> 0.2", [hex: :earmark, optional: true]}]}}
```
whatyouhide: I just pushed this change to master so it doesn't have to clutter the PR, could you rebase on master? :)

test/gettext/plural_test.exs
```diff
@@ -29,6 +29,14 @@ defmodule Gettext.PluralTest do
     assert Plural.plural("it", 100) == 1
   end
 
+  test "works for Tagalog" do
+    assert Plural.nplurals("tl") == 2
+
+    assert Plural.plural("tl", 0) == 0
+    assert Plural.plural("tl", 1) == 0
+    assert Plural.plural("tl", 100) == 1
+  end
```
whatyouhide: We don't need a test for this, no worries. We only have tests for a couple of languages just to ensure nothing major breaks, but the function clauses for Tagalog are autogenerated along a lot of other clauses so they will very very likely work :)

whatyouhide:
Great catch @Rakoth! I left a couple of comments :)

Rakoth:
Done!

whatyouhide:
Thanks a lot @Rakoth! 💟

PR #106, opened 2016-06-21
Fix plural form for numbers ends with 11 in slavic locales

Rakoth:


test/gettext/plural_test.exs
```diff
@@ -45,4 +45,9 @@ defmodule Gettext.PluralTest do
     message = ~r/unknown locale "en-us"/
     assert_raise UnknownLocaleError, message, fn -> Plural.nplurals("en-us") end
   end
+
+  test "correct plural form for X11 in slavic" do
+    assert Plural.plural("ru", 111) == 2
+    assert Plural.plural("ru", 211) == 2
+  end
```
whatyouhide: We can remove this test for the same reasons we didn't add the new test in #105 :)

whatyouhide:
Hey @Rakoth, awesome catch! I just left a tiny comment, then we're good to go, looks great to me.

test/gettext/plural_test.exs
lexmag: Not sure that this test isn't needed, we don't have test for any of `@three_forms_slavic`, that generated group may raise and we won't notice this. 😕

whatyouhide: We would need tests for every group of languages, most likely. Having one just for 3-forms slavic languages is of little use IMO 😐 I'm not against adding a test for each language family though :)

josevalim: So we can start by merging this one. :)

whatyouhide: Alright, outnumbered here 😄

PR #108, opened 2016-07-03
Reenable the "compile" task in the gettext.extract Mix task

### LINKED ISSUE #104: Task gettext.extract does not work when used last in an alias
I just stumbled upon a problem with the `mix gettext.extract` task. In our `mix.exs` file we define an alias like this:
```
      "gettext.extract": ["custom.gettext.extract", "gettext.extract"]
```
This does not work :( The code gets compiled but the POT file is never written.
However:
```
      "gettext.extract": ["gettext.extract", "custom.gettext.extract"]
```
does work.

Our custom extract task looks like this.
```
  def run(_) do
    node_env = case Application.get_env(:papyrus_psp, :environment) do
      :dev   -> "development"
      :prod  -> "production"
      :test  -> "test"
      _      -> "unknown"
    end

    case Mix.Shell.IO.cmd(command(node_env)) do
      0 -> :ok
      _ -> :error
    end
  end

  defp command(node_env) do
    "NODE_ENV=#{node_env} $(npm bin)/react-gettext-parser --output priv/gettext/react.pot 'web/static/js/**/*.{js,jsx}' '!web/static/js/i18n.js'"
  end
```

It's not really a problem for us, but maybe someone else will run into the same problem, so I wanted to document it here.

whatyouhide:
This will fix #104.

\cc @josevalim

### DIFF
--- a/lib/mix/tasks/gettext.extract.ex
+++ b/lib/mix/tasks/gettext.extract.ex
@@ -63,7 +63,14 @@ defmodule Mix.Tasks.Gettext.Extract do
 
   defp force_compile do
     Enum.map Mix.Tasks.Compile.Elixir.manifests, &make_old_if_exists/1
+    # If "compile" was never called, the reenabling is a no-op and
+    # "compile.elixir" is a no-op as well (because it wasn't reenabled after
+    # running "compile"). If "compile" was already called, then running
+    # "compile" is a no-op and running "compile.elixir" will work because we
+    # manually reenabled it.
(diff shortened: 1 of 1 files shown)


josevalim:
@whatyouhide are you sure this is enough to fix it? If you call `Mix.Task.run "compile"` in the previous task, it won't be reenable and it won't work either. I think we need to reenable both compile and compile.elixir.

whatyouhide:
@josevalim updated, :shipit:?

josevalim:
:shipit:

Issue #107, opened 2016-06-29
Alternative to missing interpolation keys error message


leighshepperson:
If you use interpolation to construct a string, and you provide it with missing keys, then it returns an error message telling you which keys are missing.

However, I need to use Gettext dynamically, and I would like it to do something a little different: If the key "key" is not present, then I would like it to return the string with place holder text "{key}".

For example, instead of this:

```elixir
Localisation.Gettext.lgettext("en", "hello_world", "score-message", %{})
{:error, "missing interpolation keys: score"}
```

I would like it to return something like this:

```elixir
Localisation.Gettext.lgettext("en", "hello_world", "score-message", %{})
{:ok, "You have scored {score} points."}
```

It would be great if the library could support this.

Unfortunately, it looks like this would be quite a big change, especially since the Interpolation module runs quite deeply in the code. 

Is there any worth in creating a branch to implement this?

whatyouhide:
Just to point out, but did you mean this call in the examples you showed:

```elixir
Localisation.Gettext.lgettext("en", "hello_world", "%{score}-message", %{})
```

? Assuming you meant that, `lgettext/4` is not public API but I'm assuming you're using it as an example to show the `{:error, _} | {:ok, _}` return value (instead of the raising that the public functions/macros do).

That said, I tend to be 👎 for this feature; I feel it's a *really* unexpected behaviour that could introduce potentially a lot and very subtle bugs (e.g., you misspell a binding and you don't see that until you look at the translated string).

What's your use case? Maybe that could help us find a better way to solve your problem. Off the top of my head, one thing I can think of doing is exposing some function in `Gettext.Interpolation` that returns all the bindings in a given string (IIRC we have something like that already, it would just be a matter of making it public). But let's look at the use case first, the less stuff we add, the easier it will be to maintain Gettext :)

maufl:
As I said in #109 I don't like this behaviour either. I understand why not raising an error might not be a good idea. On the other hand the translations are not always maintained by the developers and are not usually all covered by tests. 
So if a translator makes a mistake in a binding name, we will get a runtime error. :( 

As an example, we have a website with multiple locales. Now if the Norwegian translation contains an typo in the binding, we will never know and our customer might get an 500 instead of simply a warning like `<missing placeholer>`. I guess it is unlikely that they will report a 500 and more likely that they report a broken translation.

whatyouhide:
[...]

PR #109, opened 2016-07-04
Allow recursive lookup of interpolation keys

maufl:
Hi guys, we are using this module in our Phoenix app, and we have some nested models. In some places we want to interpolate information from one or more models into translatable strings. It would be very handy if we did not have to build a intermediate map but could just interpolate like this:
```
gettext("Welcome %{user.first_name} %{user.last_name} to your project %{project.name}", %{user: user, project: project})
```

My pull request would allow just that. What's your opinion on this feature?
Also, what's missing from my pull request? (documentation, more tests)

### DIFF
--- a/lib/gettext/interpolation.ex
+++ b/lib/gettext/interpolation.ex
@@ -102,14 +102,36 @@ defmodule Gettext.Interpolation do
     segments = to_interpolatable(str)
     keys     = keys(segments)
 
-    if keys -- Map.keys(bindings) != [] do
+    if keys -- map_recursive_keys(bindings) != [] do
       {:error, missing_interpolation_keys(bindings, keys)}
     else
       interpolated = Enum.map_join segments, "", fn
-        key when is_atom(key) -> Map.fetch!(bindings, key)
+        key when is_atom(key) -> map_recursive_fetch!(bindings, key)
         other                 -> other
       end
       {:ok, interpolated}
     end
   end
+
+  defp map_recursive_fetch!(map, key) do
+    key
+    |> Atom.to_string()
+    |> String.split(".")
+    |> Enum.reduce(map, fn(k, m) -> Map.fetch!(m, String.to_atom(k)) end)
+  end
+
+  defp map_recursive_keys(map, prefix \\ nil) do
+    map
+    |> Map.keys()
+    |> Enum.map(fn key ->
+      case Map.fetch!(map, key) do
+        %{} = value -> map_recursive_keys(value, :"#{dot_join([prefix, key])}")
+        _ -> :"#{dot_join([prefix, key])}"
+      end
+    end)
+    |> Enum.to_list()
+    |> List.flatten()
+  end
+
+  defp dot_join(list), do: list |> Enum.filter(&(&1)) |> Enum.join(".")
 end
--- a/test/gettext/interpolation_test.exs
+++ b/test/gettext/interpolation_test.exs
@@ -47,6 +47,12 @@ defmodule Gettext.InterpolationTest do
            == {:ok, "Hello Alex. Goodbye Alex"}
   end
 
+  test "interpolate/2: nested maps" do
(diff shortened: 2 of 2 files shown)


josevalim:
I actually like this feature. What about you @whatyouhide? However, I think you have implemented it only for dynamic lookups and not compiled ones. Should we add a test for a compiled lookup?

Also, I wouldn't do map_recursive_keys. I think we should check the key presence only at the first level. Everything else will fail with a Map.fetch! anyway.

whatyouhide:
I am still not sure about this feature; I can see the benefit it brings by doing

```elixir
fox = %{color: "brown"}
gettext "The %{fox.color} jumps ...", fox: fox
```

over

```elixir
gettext "The %{fox_color} fox ...", fox_color: fox.color
```

but I am also worried that it may make things harder to grasp and maybe debug. In the current implementation, you're responsible of possibly fetching keys from your maps or whatever you need to do to fetch the bindings, but here we'd be in charge of decisions like: do we use `Map.fetch!/2` on the bindings and let that raise? should we raise a dedicated error message? and so on. Also, maybe you'll change the map format and forget to update the binding in the Gettext strings, and problems like this (that may or may not be similar without this feature, but just to bring them up).

Also, this would complicate the validity of what can go inside `%{}` because not it can contain `.`s, but not at the beginning/end, and so on (so the `String.split/2` in the proposed implementation would not be enough I'm afraid). Just a thing to consider, not saying this is inherently bad :)

As far as the implementation goes, I agree that we can drop `map_recursive_keys` and just check the top-level of the map bindings, and raise if keys are missing when we go ahead and fetch them later.

A final note implementation-wise: we should probably change `to_interpolatable` to return a list of atoms instead of an atom for a binding (a-la `get_in/2` and friends) so that we don't have to keep atoms like `:"foo.bar.baz"` and then convert them to strings and split them, but we can have:

```elixir
iex> to_interpolatable("The %{fox.color} jumps over the %{what}")
["The ", [:fox, :color], " jumps over the ", [:what]]
```

I *think* (haven't looked too deeply) this would make things clearer and maybe simpler.

But the main question remains, shall we support this or let the user do the binding fetching? :)

maufl:
Another thought I just had with regard to missing keys in the bindings and Map.fetch!: If whoever translates a string mistypes a placeholder name, then we will have a runtime error. I don't really like this behaviour. I understand that not getting an error when a key is missing is also undesirable but I would hate to have a translation error result in a 500 on a website.

josevalim:
That's a good point. Can you please open up a separate discussion for that?

On Monday, July 4, 2016, Felix Maurer <notifications@github.com> wrote:

> Another thought I just had with regard to missing keys in the bindings and
> Map.fetch!: If whoever translates a string mistypes a placeholder name,
> then we will have a runtime error. I don't really like this behaviour. I
> understand that not getting an error when a key is missing is also
> undesirable but I would hate to have a translation error result in a 500 on
> a website.
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/elixir-lang/gettext/pull/109#issuecomment-230271322>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAAlbmpoWCkusrDEwxbWyyeWSfspZdO3ks5qSPEJgaJpZM4JEM04>
> .
>


-- 


maufl:
I think it is already discussed in #107. The more I think about it the less I'm sure whether this is a good idea and what the right solution would look like.

maufl:
I did not have a lot of use cases for this recently and I'm still unsure how to handle the difference between a Map key with a dot or a subkey. So I would close this PR for now if you don't mind.

whatyouhide:
I am definitely 👍 on closing if we don't have many use cases because the current code is simpler, that's all :)

We can revisit later if we feel like it, thanks anyways @maufl! 💟

PR #111, opened 2016-07-19
Read the config both from "use Gettext" and Mix.Config

whatyouhide:
The two will be merged, giving precedence to the configuration passed to `use Gettext`.

\cc @josevalim, this looks pretty good to me, but I would love another pair of eyes on docs especially :)


josevalim:
:shipit:

PR #113, opened 2016-07-27
Add missing bindings handler

maufl:
This is far from finished but I struggle with the meta programming, so I just open the discussion here.
When running the tests, I get an error that I do not understand.
```
  1) test lngettext/6: default handle_missing_binding preserves key (GettextTest)
     test/gettext_test.exs:189
     ** (ArgumentError) argument error
     stacktrace:
       :erlang.bit_size(6)
       test/gettext_test.exs:1: anonymous fn/2 in GettextTest.Translator.lngettext/6
       (elixir) lib/enum.ex:1473: Enum."-reduce/3-lists^foldl/2-0-"/3
       test/gettext_test.exs:1: GettextTest.Translator.lngettext/6
       test/gettext_test.exs:196
```

### DIFF
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -444,6 +444,19 @@ defmodule Gettext do
     quote do
       @gettext_opts unquote(opts)
       @before_compile Gettext.Compiler
+
+      @doc """
+      Default handling for missing bindings.
+
+      If a translated `msgid` does contain a binding that is missing, this method will be called.
+      By default it only logs the error but it can be overwritten.
+      The return value is used instead of the missing binding.
+      """
+      @spec handle_missing_binding(atom, binary, binary) :: binary
+      def handle_missing_binding(binding, original_string, locale) do
+        "%{#{binding}}"
+      end
+      defoverridable handle_missing_binding: 3
     end
   end
 
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -450,9 +450,24 @@ defmodule Gettext do
       @doc """
       Default handling for missing bindings.
 
-      If a translated `msgid` does contain a binding that is missing, this method will be called.
-      By default it only logs the error but it can be overwritten.
-      The return value is used instead of the missing binding.
+      This function is called for every missing binding found in a translation.
+      It takes the `binding` (as an atom), the original translation, and the current locale.
+      The return value of this function is used as replacement for the missing binding.
+      For example, if something like this is called:
+
+          MyApp.Gettext.gettext("Hello %{name}, welcome to %{country}", name: "Jane", country: "Italy")
+
+      and our `it/LC_MESSAGES/default.po` looks like this:
+
+          msgid "Hello %{name}, welcome to %{country}"
+          msgstr "Ciao %{name}, benvenuto in %{cowntry}" # (typo)
+
+      then Gettext will call:
+
+          MyApp.Gettext.handle_missing_binding(:cowntry, "Ciao %{name}, benvenuto in %{cowntry}", "it")
+
+      The default implementation for this function warns about the missing binding and returns `%{binding}` (`%{cowntry}` in the example).
+      This function can be overridden.
       """
       @spec handle_missing_binding(atom, binary, binary) :: binary
       def handle_missing_binding(binding, original_string, locale) do
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -442,6 +442,8 @@ defmodule Gettext do
   @doc false
   defmacro __using__(opts) do
     quote do
+      require Logger
+
       @gettext_opts unquote(opts)
       @before_compile Gettext.Compiler
 
@@ -454,6 +456,7 @@ defmodule Gettext do
       """
       @spec handle_missing_binding(atom, binary, binary) :: binary
       def handle_missing_binding(binding, original_string, locale) do
+        Logger.warn("The key \"#{binding}\" for the translation \"#{original_string}\" (locale #{locale}) is missing from the bindings.")
         "%{#{binding}}"
       end
       defoverridable handle_missing_binding: 3
--- a/lib/gettext/compiler.ex
+++ b/lib/gettext/compiler.ex
@@ -129,25 +129,27 @@ defmodule Gettext.Compiler do
   @spec dynamic_clauses() :: Macro.t
   def dynamic_clauses do
     quote do
-      def lgettext(_locale, domain, msgid, bindings) do
+      def lgettext(locale, domain, msgid, bindings) do
         Gettext.Compiler.warn_if_domain_contains_slashes(domain)
 
-        case Gettext.Interpolation.interpolate(msgid, bindings) do
-          {:ok, interpolated} -> {:default, interpolated}
-          {:error, _} = error -> error
-        end
+        {:default, interpolate(msgid, bindings, locale)}
       end
 
-      def lngettext(_locale, domain, msgid, msgid_plural, n, bindings) do
+      def lngettext(locale, domain, msgid, msgid_plural, n, bindings) do
         Gettext.Compiler.warn_if_domain_contains_slashes(domain)
 
         str      = if n == 1, do: msgid, else: msgid_plural
         bindings = Map.put(bindings, :count, n)
 
-        case Gettext.Interpolation.interpolate(str, bindings) do
-          {:ok, interpolated} -> {:default, interpolated}
-          {:error, _} = error -> error
-        end
+        {:default, interpolate(str, bindings, locale)}
+      end
+
+      defp interpolate(str, bindings, locale) do
+        Gettext.Interpolation.to_interpolatable(str)
+        |> Enum.map_join("", fn
+          segment when is_atom(segment) -> Map.get_lazy(bindings, segment, fn -> handle_missing_binding(segment, str, locale) end)
+          segment                       -> segment
+        end)
       end
     end
   end
@@ -285,12 +287,10 @@ defmodule Gettext.Compiler do
         unquote(match) ->
           {:ok, unquote(interpolation)}
         %{} ->
-          translation = unquote(interpolatable)
-          |> Enum.map(fn
-            segment when is_binary(segment) -> segment
+          translation = Enum.map_join(unquote(interpolatable) , "", fn
             segment when is_atom(segment) -> Map.get_lazy(var!(bindings), segment, fn -> handle_missing_binding(segment, unquote(str), unquote(locale)) end)
+            segment -> segment
           end)
-          |> Enum.reduce("", fn segment, acc -> acc <> segment end)
           {:ok, translation}
         _ ->
           raise(Gettext.Error, "Bindings must be a map")
--- a/lib/gettext/compiler.ex
+++ b/lib/gettext/compiler.ex
@@ -147,8 +147,10 @@ defmodule Gettext.Compiler do
       defp interpolate(str, bindings, locale) do
         Gettext.Interpolation.to_interpolatable(str)
         |> Enum.map_join("", fn
-          segment when is_atom(segment) -> Map.get_lazy(bindings, segment, fn -> handle_missing_binding(segment, str, locale) end)
-          segment                       -> segment
+          segment when is_atom(segment) ->
+            Map.get_lazy(bindings, segment, fn -> handle_missing_binding(segment, str, locale) end)
+          segment ->
+            segment
         end)
       end
     end
--- a/lib/gettext/compiler.ex
+++ b/lib/gettext/compiler.ex
@@ -304,8 +304,6 @@ defmodule Gettext.Compiler do
             segment -> segment
           end)
           {:ok, translation}
-        _ ->
-          raise(Gettext.Error, "Bindings must be a map")
       end
     end
   end
--- a/lib/gettext/compiler.ex
+++ b/lib/gettext/compiler.ex
@@ -146,12 +146,7 @@ defmodule Gettext.Compiler do
 
       defp interpolate(str, bindings, locale) do
         Gettext.Interpolation.to_interpolatable(str)
-        |> Enum.map_join("", fn
-          segment when is_atom(segment) ->
-            Map.get_lazy(bindings, segment, fn -> handle_missing_binding(segment, str, locale) end)
-          segment ->
-            segment
-        end)
+        |> Gettext.Interpolation.interpolate(bindings, str, locale, &__MODULE__.handle_missing_binding/3)
       end
     end
   end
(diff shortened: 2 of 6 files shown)


maufl:
Also I get a lot of 
```
test/gettext_test.exs:9: warning: this clause cannot match because a previous clause at line 9 always matches
```
and I understand why it happens but my Macro foo is not strong enough yet to come up with a solution.

maufl:
I rewrote the dynamic clauses. It's likely that I overlook something, let me know what you think.

lib/gettext/compiler.ex
```diff
@@ -281,18 +283,24 @@ defmodule Gettext.Compiler do
...
+        %{} ->
+          translation = Enum.map_join(unquote(interpolatable) , "", fn
+            segment when is_atom(segment) -> Map.get_lazy(var!(bindings), segment, fn -> handle_missing_binding(segment, unquote(str), unquote(locale)) end)
+            segment -> segment
+          end)
+          {:ok, translation}
         _ ->
-          keys = unquote(keys)
-          {:error, Gettext.Interpolation.missing_interpolation_keys(var!(bindings), keys)}
+          raise(Gettext.Error, "Bindings must be a map")
```
whatyouhide: Let's fix one thing at a time starting with the compile-time clauses. The reason you get a warning that "this clause will always match" is that take the example where `str` is something like `"hello world"`, with no bindings. What this will compile to is something like:

```elixir
case bindings do
  %{} ->
    {:ok, "hello world"}
  %{} -> 
    ...
  _ ->
    raise ...
end
```

and as you can see the first two clauses are equal. I think the right thing to do is handle the cause where the string has no interpolations separately; this can be done in different way, for example:

```elixir
    if keys == [] do
      quote do
        {:ok, unquote(str)}
      end
    else
      quote do
      case var!(bindings) do
          unquote(match) ->
            {:ok, unquote(interpolation)}
          %{} ->
            translation =
              Enum.map_join(unquote(interpolatable), "", fn
                segment when is_binary(segment) ->
                  segment
                binding when is_atom(binding) ->
                  Map.get_lazy(var!(bindings), binding, fn -> handle_missing_binding(binding, unquote(str), unquote(locale)) end)
              end)
            {:ok, translation}
          _ ->
            raise(Gettext.Error, "Bindings must be a map")
        end
      end
    end
```

This works but the only problem is that when there are no interpolations the compiler will complain that `bindings` is unused, and it's right because our final code will look something like:

```elixir
bindings = ...
{:ok, "hello world"}
```

instead of 

```elixir
bindings = ...
case bindings do
  ...
end
```

One solution I found to work is to do something like:

```elixir
    if keys == [] do
      quote do
        _ = var!(bindings)
        {:ok, unquote(str)}
      end
```

and that will silence the warnings. However there may be a better way close to something like:

```elixir
case var!(bindings) do
  %{} when unquote(keys) == [] ->
    {:ok, unquote(str)}
  ...
end
```

The problem with this is that the compiler will complain about always-matching clauses because this will compile to `%{} when [] == []`. @josevalim can I ask your help here? Do you have a better solution in mind than the long `if` with `_ = var!(bindings)` I showed above?

maufl:
My latest update seems to remove all warnings.

lib/gettext/compiler.ex
whatyouhide: I think we can remove this raising because we check that the bindings are a map with guards in the functions, so they should never arrive here as a non-map.

lib/gettext/compiler.ex
```diff
@@ -281,18 +283,34 @@ defmodule Gettext.Compiler do
...
   # the quoted code that is returned.
-  defp compile_interpolation(str) do
-    keys          = Interpolation.keys(str)
-    match         = compile_interpolation_match(keys)
-    interpolation = compile_interpolatable_string(str)
+  defp compile_interpolation(str, locale) do
+    compile_interpolation(str, locale, Interpolation.keys(str))
+  end
+
+  defp compile_interpolation(str, _locale, []=_keys) do
```
whatyouhide: We use spaces around `=`: `...(str, _locale, [] = _keys)` :)

lib/gettext/compiler.ex
```diff
@@ -281,18 +283,34 @@ defmodule Gettext.Compiler do
...
+    match          = compile_interpolation_match(keys)
+    interpolation  = compile_interpolatable_string(str)
+    interpolatable = Interpolation.to_interpolatable(str)
 
     quote do
       case var!(bindings) do
         unquote(match) ->
           {:ok, unquote(interpolation)}
+        %{} ->
+          translation = Enum.map_join(unquote(interpolatable) , "", fn
```
whatyouhide: There's a space before a comma here.

whatyouhide:
@josevalim can I ask your opinion on this PR? It looks ok but I'm not 100% sure this is the way to go. Also if we do this this way I think we're not using `Gettext.Interpolation.interpolate/2` anymore, correct @maufl?

lib/gettext.ex
```diff
@@ -430,6 +430,19 @@ defmodule Gettext do
...
+      @doc """
+      Default handling for missing bindings.
+
+      If a translated `msgid` does contain a binding that is missing, this method will be called.
+      By default it only logs the error but it can be overwritten.
+      The return value is used instead of the missing binding.
+      """
+      @spec handle_missing_binding(atom, binary, binary) :: binary
+      def handle_missing_binding(binding, original_string, locale) do
+        "%{#{binding}}"
```
whatyouhide: We would need to warn here by default as well.

maufl: Yes, I forgot to ask. I tried using Logger but then noticed that Gettext currently does not use Logger and it would have to be started as application as far as I understand. So I was not sure whether I should add the logging.

maufl:
Yes, `Gettext.Interpolation.interpolate/2` is not used anymore as far as I can see.

lib/gettext.ex
whatyouhide: We can add `:logger` as an application, that would be ok IMO, or use `IO.puts(:stderr, "warning: ...")` and be more manual. Wdyt @josevalim?

lib/gettext/compiler.ex
```diff
@@ -132,25 +132,27 @@ defmodule Gettext.Compiler do
...
-          {:ok, interpolated} -> {:default, interpolated}
-          {:error, _} = error -> error
-        end
+        {:default, interpolate(str, bindings, locale)}
+      end
+
+      defp interpolate(str, bindings, locale) do
+        Gettext.Interpolation.to_interpolatable(str)
+        |> Enum.map_join("", fn
+          segment when is_atom(segment) -> Map.get_lazy(bindings, segment, fn -> handle_missing_binding(segment, str, locale) end)
```
lexmag: This is a quite long line, I think if we put right side of `->` on the next line it will improve readability.

whatyouhide: Agreed, that would definitely improve readability :)

maufl: Agreed :)

whatyouhide:
@josevalim do you have a quick opinion on this before ElixirConf starts? :)

lib/gettext.ex
josevalim: :+1: for Logger with warning level.

whatyouhide:
Hey @maufl :), do you have time to work on this? I have a few more comments :)

maufl:
I don't have time but I can make time ;)

whatyouhide:
@maufl I can pick this PR up if you don't feel like spending time on it, don't hesitate to say so, it's absolutely fine :D In the meantime, I'll leave those comments so if you decide to work on it we can move forward 💟

maufl:
@whatyouhide I'll keep on working on this PR :) Just leave the comments

lib/gettext.ex
```diff
@@ -430,6 +430,19 @@ defmodule Gettext do
       @gettext_opts unquote(opts)
       @before_compile Gettext.Compiler
       unquote(Gettext.Compiler.signatures)
+
+      @doc """
+      Default handling for missing bindings.
+
+      If a translated `msgid` does contain a binding that is missing, this method will be called.
```
whatyouhide: Let's change the documentation to:

```
This function is called for every missing binding found in a translation. It takes the `binding` (as an atom), the original translation, and the current locale. The return value of this function is used as replacement for the missing binding. For example, if something like this is called:

    MyApp.Gettext.gettext("Hello %{name}, welcome to %{country}", name: "Jane", country: "Italy")

and our `it/LC_MESSAGES/default.po` looks like this:

    msgid "Hello %{name}, welcome to %{country}"
    msgstr "Ciao %{name}, benvenuto in %{cowntry}" # (typo)
   
then Gettext will call:

    MyApp.Gettext.handle_missing_binding(:cowntry, "Ciao %{name}, benvenuto in %{cowntry}", "it")

The default implementation for this function warns about the missing binding and returns `%{binding}` (`%{cowntry}` in the example). This function can be overridden.
```

Wdyt? Btw, double check I got the docs right, always helpful to have two pair of eyes 😄

maufl: Looks good to me :) I updated the documentation and added the logging

lib/gettext/compiler.ex
```diff
@@ -274,18 +278,34 @@ defmodule Gettext.Compiler do
...
 
     quote do
       case var!(bindings) do
         unquote(match) ->
           {:ok, unquote(interpolation)}
+        %{} ->
+          translation = Enum.map_join(unquote(interpolatable), "", fn
+            segment when is_atom(segment) -> Map.get_lazy(var!(bindings), segment, fn -> handle_missing_binding(segment, unquote(str), unquote(locale)) end)
+            segment -> segment
+          end)
```
whatyouhide: This logic is exactly the same as the one above. We should likely extract this, possibly to `Gettext.Interpolation`.

lib/gettext/compiler.ex
whatyouhide: @maufl maybe you missed this last comment here, but let's drop the handling of non-maps. :)

test/gettext_test.exs
```diff
@@ -215,10 +224,10 @@ defmodule GettextTest do
     assert Translator.dgettext("interpolations", "Hello %{name}", keys)
            == "Ciao Jim"
 
-    msg = "missing interpolation keys: name"
-    assert_raise Gettext.Error, msg, fn ->
-      Translator.dgettext("interpolations", "Hello %{name}")
-    end
+    assert capture_log(fn ->
+      assert Translator.dgettext("interpolations", "Hello %{name}")
```
whatyouhide: Let's put this assertion on one line, it's short enough :)

test/gettext_test.exs
```diff
@@ -215,10 +224,10 @@ defmodule GettextTest do
...
+      == "Ciao %{name}"
+    end) =~ ~s[The key "name" for the translation "Ciao %{name}" (locale it) is missing from the bindings.]
```
whatyouhide: We could refactor this to be a bit cleaner by doing:

```elixir
log = capture_log(fn ->
  assert ...
end)

assert log =~ ~s[...]
```

wdyt?

lib/gettext/compiler.ex
maufl: How would you do that? The `handle_missing_binding` method is only defined in `Gettext` and not in `Gettext.Interpolation`.

test/gettext_test.exs
```diff
@@ -22,6 +22,7 @@ defmodule GettextTest do
   use ExUnit.Case
```
lexmag: It should be `async: false` since we have `import ExUnit.Capture*`.

whatyouhide: `:async` defaults to `false` luckily 😄

maufl: Is `async: false` not the default?

maufl:
@whatyouhide I implemented your suggestions :)

lib/gettext/compiler.ex
whatyouhide: Since `handle_missing_binding/3` is public, you can pass the `__MODULE__` to the new function in `Gettext.Interpolation` and that can call `module.handle_missing_binding/3`.

test/gettext_test.exs
lexmag: Yeah. :bowtie:

maufl:
@whatyouhide Sorry, for the delay. I extracted the interpolation logic. I pass the function instead of the module, that felt better to me. But I can pass the module if you think it's cleaner.

I was unable to spread the example in the documentation over multiple lines, the compiler kept complaining about the indentation. :(

lib/gettext/interpolation.ex
```diff
@@ -33,22 +33,25 @@ defmodule Gettext.Interpolation do
...
-  Returns an error message which tells which keys in `required` don't appear in
-  `bindings`.
+  This function takes an interpolatable list returned from `to_interpolatable/1` and bindings
+  and returns the interpolated string. If it encounters an atom that should be interpolated
+  but is missing from the bindings, it will call the provided `handle_missing_binding` function.
+  The callback will be called with the missing binding, the original string and the locale.
+  See also the default implementation in `Gettext`.
 
-  ## Examples
+  ## Example
```
whatyouhide: This should remain `Examples`.

lib/gettext/interpolation.ex
```diff
@@ -33,22 +33,25 @@ defmodule Gettext.Interpolation do
...
 
-      iex> Gettext.Interpolation.missing_interpolation_keys %{foo: 1}, [:foo, :bar, :baz]
-      "missing interpolation keys: bar, baz"
+      iex> Gettext.Interpolation.interpolate(["Hello ", :name, ", you have ", :count, " unread messages"], %{ :name => "José", :count => 3 }, "Hello %{name}, you have %{count} unread messages", "en_GB", &__MODULE__.handle_missing_binding/3)
```
whatyouhide: We should strive to avoid a huge line here. Use something like this:

```elixir
iex> string = ["Hello ", :name, ", you have ", :count, " unread messages"]
iex> msgid = "Hello %{name}, you have %{count} unread messages"
iex> handler = fn(binding, _str, _locale) -> Atom.to_string(binding) end
iex> Gettext.Interpolation.interpolate(string, %{name: "José", count: 3}, msgid, "en_GB", handler)
"Hello José, you have 3 unread messages"
```

maufl: Ah, I could have thought of that myself.

maufl:
I cannot see the reason why the tests fail. Locally everything works.

whatyouhide:
Ah, no worries about the tests, I can look into it after merging, it's only failing for one Erlang/Elixir combo.

josevalim:
I had an idea for this. What if `handle_missing_bindings` accept the `Gettext.InterpolationError` struct? This way it should be easy to choose between warning, erroring or doing nothing. I will take a look at this soon.

josevalim:
Thank you! I have rebased and made it part of #129! 

Issue #115, opened 2016-08-16
Dialyzer warnings with OTP 19


aaronjensen:
Brand new phoenix app created with `mix phoenix.new` on OTP 19 has these warnings.

Here's a repro app: https://github.com/aaronjensen/dialyzer_repro

Irrelevant warnings are omitted. Originally reported at https://github.com/phoenixframework/phoenix/issues/1872

```
Starting Dialyzer
dialyzer --no_check_plt --plt /Users/aaronjensen/.dialyxir_core_19_1.3.2.plt -Wunmatched_returns -Werror_handling -Wrace_conditions -Wunderspecs /Users/aaronjensen/Source/dialyzer_repro/_build/dev/lib/dialyzer_repro/ebin
  Proceeding with analysis...
gettext.ex:1: Function lngettext/6 has no local return
gettext.ex:1: The call 'Elixir.Gettext.Interpolation':interpolate(_@7::any(),_@8::#{'count':=_, _=>_}) breaks the contract (binary(),#{}) -> {'ok',binary()} | {'error',binary()}
gettext.ex:23: Function lngettext/5 has no local return
 done in 0m3.37s
done (warnings were emitted)
```

whatyouhide:
@josevalim is this a dialyzer bug as well or should I look into it? :)

josevalim:
They don't seem to be dialyzer bugs. They look like a missing spec for lngettext functions (they are required when there is no local return) and the other one looks like dialyzer thinks we are passing more than string to interpolate. Both can probably be fixed by specifying specs for lngettext.

whatyouhide:
Got it, gonna do this then :)

whatyouhide:
@aaronjensen I just pushed a couple of commits that may fix part of this, can you give it a try again and report if you get any more warnings? :)

aaronjensen:
Different ones now:

```
gettext.ex:1: The call 'Elixir.Gettext.Interpolation':interpolate(_@7::any(),_@8::#{'count':=_, _=>_}) breaks the contract (binary(),#{}) -> {'ok',binary()} | {'error',binary()}
gettext.ex:23: The specification for 'Elixir.DialyzerRepro.Gettext':lgettext/4 states that the function might also return {'ok',binary()} but the inferred return is {'default',binary()} | {'error',<<_:64,_:_*8>>}
gettext.ex:23: Function lngettext/5 has no local return
[...]

PR #118, opened 2016-08-22
Only check for the type of bindings in Gettext.*gettext functions

whatyouhide:
`l*gettext` functions generated in backends will now assume bindings are a map and the only conversion from keyword list to map will be performed in `Gettext.*gettext` functions. These functions are used by the `*gettext` macros generated in each backend as well, so each case will be covered.

@josevalim wdyt of this? We were using `bindings \\ %{}` in `l*gettext` functions but those are never called directly (except in tests), so I think being stricter makes sense there. Also typespecs should be a bit clearer this way (related to #115 which I'm trying to fix).


josevalim:
:+1:

PR #119, opened 2016-08-23
Refactor some function generation in Gettext.Compiler

whatyouhide:
We basically pointlessly used `Enum.reduce/3` instead of good old `Enum.flat_map/2` :) The only thing that changes now is the order of the generated clauses, but that's ok since they're all different clauses.


josevalim:
@whatyouhide you should totally push most of those changes directly. :)

josevalim:
:shipit:

whatyouhide:
@josevalim yes, sorry about that, I mostly opened this as a PR for myself so that I could play with it for a short amount of time to ensure I didn't break anything 😛 I will say so in the PR next time!

PR #124, opened 2016-09-19
Pop test project on test exit.

### LINKED ISSUE #123: "mix test --stale" doesn't work with Gettext
I have to investigate why. It basically just runs as `mix test` without caching tests or even printing anything that `mix test --stale` usually prints. @antipax if you have any quick idea on why this could be, feel free to chime in :)

antipax:
Fixes #123.


whatyouhide:
Looks good @antipax, thanks a lot. For reference, what was happening? :)

antipax:
By setting the current mix project to `:my_project`, at the end of the test suite the mix project would be `:my_project` and thus the `mix test --stale` manifest would be written to the `my_project` build dir rather than the `gettext` one. Thus on subsequent runs when `mix test --stale` looked for the manifest it could not be found. :)

whatyouhide:
Okay, that makes sense, thanks a lot for the quick fix! 💟

PR #127, opened 2016-10-22
Ensure extractions are tracked to proper app, closes #126

### LINKED ISSUE #126: Running mix gettext.extract in umbrella updates wrong app
Hi, I'm moving our project that uses gettext into an umbrella and I noticed that running `mix gettext.extract` from umbrella root updates the wrong app.

In a sample app, I install gettext in app `foo`, but extracting from the umbrella root adds files to the `bar` app: https://github.com/wojtekmach/gettext_umbrella/commit/f43292c2c014ab86b327b369576d03c91091809a. When I first run `cd apps/foo && mix gettext.extract` it does the correct thing, and then running extract again from root does nothing which is good too.

Btw, are there any recommendations for using gettext in an umbrella? E.g.:

1. a single `MyProject.Gettext` module that some apps depend on, and a single `priv/gettext` directory.
2. `Foo.Gettext`, `Bar.Gettext` etc, each with it's own `priv/gettext` (need to remember to call `Gettext.put_locale(Foo.Gettext, locale) && Gettext.put_locale(Bar.Gettext, locale)` in function that changes locale for the app though)
3. It depends :-)

Thanks!

josevalim:


josevalim:
Closes #126.

@whatyouhide please see the second commit and merge if it looks good to go. :D

whatyouhide:
Looks great @josevalim, thanks for looking into this 💟

PR #129, opened 2016-10-22
Add missing bindings handler

### LINKED ISSUE #107: Alternative to missing interpolation keys error message

josevalim:
Incorporates #113, closes #107.

### DIFF
--- a/lib/gettext/backend.ex
+++ b/lib/gettext/backend.ex
@@ -10,9 +10,10 @@ defmodule Gettext.Backend do
   @doc """
   Default handling for missing bindings.
 
-  This function is called when there are missing bindings in a translation.
-  It takes the `Gettext.MissingBindingsError` struct and the incomplete translation.
-  The current backend, locale, domain and msgid can be found in the error struct.
+  This function is called when there are missing bindings in a translation. It
+  takes a `Gettext.MissingBindingsError` struct and the translation with the
+  wrong bindings left as is with the `%{}` syntax.
+
   For example, if something like this is called:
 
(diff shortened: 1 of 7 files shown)


lib/gettext/backend.ex
```diff
@@ -6,6 +6,39 @@ defmodule Gettext.Backend do
...
+
+  The default implementation for this function uses `Logger.error/1` to warn
+  about the missing binding and returns the incomplete message.
+
+  This function can be overridden to raise, for example:
+
+      def handle_missing_bindings(exception, _incomplete) do
+        raise exception
+      end
+  """
```
whatyouhide: In the docs we should mention that the value returned by `handle_missing_bindings/2` is used as the translated string, and that the default implementation just returns the string with the missing bindings left untouched.

PR #130, opened 2016-10-22
Use Logger.error/1 for runtime warnings

### LINKED ISSUE #128: Use Logger.error/1 to warn users about slashes in domains
Right now we use hand-rolled `IO.puts(:stderr, ...)`, but `Logger` would be better.

whatyouhide:
Instead of `IO.puts(:stderr, msg)`.

Closes #128.


josevalim:

PR #138, opened 2016-11-30
Replace usage of String.*strip functions

whatyouhide:
They are soft deprecated as of Elixir 1.4 and will be hard deprecated by Elixir 1.5. The trim/replace functions that we use now are available since Elixir 1.2 so we are fine.

Let's see if Travis agrees before merging this, that's why I didn't push to master 😛


whatyouhide:
Nope, had things mixed up: the `trim_*` functions are available on 1.3 and only soft-deprecated in 1.4, so we shall remove them only when we depend on ~1.3. :)

PR #140, opened 2016-12-10
Define *gettext_noop macros that mark translations for extraction

### LINKED ISSUE #131: Add a way of marking a string for translation
I would like to propose a feature that is also found in the [ruby gettext gem](https://github.com/ruby-gettext/gettext):

`= N_("Apple") which does not translate, but tells the parser: "Apple" needs translation.`

This would be useful for the `mix gettext.extract` task and add "Apple" as msgid to the po files.

I would especially like to use such a feature in my Ecto validations. Currently there isn't really a nice way to add translatable error messages without manually adding them to the .po files. 
Using gettext from within the model ( `add_error(changeset, :field, gettext("Some error"))`) in a standard Phoenix project would result in double translation, as the generated ErrorHelpers will translate the already translated error message again.

whatyouhide:
Would close #131.

\cc @manukall

@josevalim I went with this approach in `*ngettext_noop` where `{msgid, msgid_plural}` is returned because that is what is often used to represent a plural translation (in Ecto for example, I think right?) and because the Wordpress implementation of these functions (one of the few I found) does the same (returns every info you pass to it basically). The Python implementation seems to ask for `n` as well and does the English pluralization rule, but of the two I think I prefer the tuple one. Wdyt?

### DIFF
--- a/lib/gettext/backend.ex
+++ b/lib/gettext/backend.ex
@@ -133,10 +133,9 @@ defmodule Gettext.Backend do
   @doc """
   Marks the given translation for extraction and returns it unchanged.
 
-  This macro can be used to mark a translation for extraction so that it is
-  extracted when `mix gettext.extract` is run. The return value is the given
-  string, so that this macro can be used seamlessly in place of the string to
-  extract.
+  This macro can be used to mark a translation for extraction when `mix
+  gettext.extract` is run. The return value is the given string, so that this
+  macro can be used seamlessly in place of the string to extract.
 
   ## Examples
 
@@ -147,16 +146,17 @@ defmodule Gettext.Backend do
   @macrocallback dgettext_noop(domain :: String.t, msgid :: String.t) :: Macro.t
 
   @doc """
-  Same as `dgettext_noop("default", msgid).
+  Same as `dgettext_noop("default", msgid)`.
   """
   @macrocallback gettext_noop(msgid :: String.t) :: Macro.t
 
   @doc """
-  Marks the given translation for extract and returns `{msgid, msgid_plural}`.
+  Marks the given translation for extraction and returns
+  `{msgid, msgid_plural}`.
 
-  This macro can be used to mark a translation for extract so that it is
-  extracted when the `mix gettext.extract` Mix task is run. The return value of
-  this macro is `{msgid, msgid_plural}`.
+  This macro can be used to mark a translation for extraction when `mix
+  gettext.extract` is run. The return value of this macro is `{msgid,
+  msgid_plural}`.
 
   ## Examples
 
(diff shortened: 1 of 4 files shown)


lib/gettext/backend.ex
```diff
@@ -129,4 +129,50 @@ defmodule Gettext.Backend do
...
+
+  @doc """
+  Same as `dgettext_noop("default", msgid).
+  """
+  @macrocallback gettext_noop(msgid :: String.t) :: Macro.t
+
+  @doc """
+  Marks the given translation for extract and returns `{msgid, msgid_plural}`.
+
+  This macro can be used to mark a translation for extract so that it is
```
josevalim: "extraction"?

lib/gettext/backend.ex
```diff
@@ -129,4 +129,50 @@ defmodule Gettext.Backend do
   @macrocallback ngettext(msgid :: String.t,
                           msgid_plural :: String.t,
                           n :: Macro.t) :: Macro.t
+
+  @doc """
+  Marks the given translation for extraction and returns it unchanged.
+
+  This macro can be used to mark a translation for extraction so that it is
+  extracted when `mix gettext.extract` is run. The return value is the given
```
josevalim: This explanation and the ones below seem a bit redundant "mark a translation for extraction so that it is extracted when `mix gettext.extract` is run". Maybe "mark a transaction for extractions when `mix gettext.extract` runs"?

whatyouhide: Good point, it sounded indeed pretty weird now that I read it again :)

josevalim:
LGTM!

manukall:
this looks great. thanks @whatyouhide .

regarding the return value of `dngettext_noop`: i, too, am not sure about the best approach here. our use case was ecto validations, where in the schema we only want to mark the error string for extraction and then in the view have it translated. returning the tuple from `dngettext_noop` won't work with the phoenix generated `ErrorHelpers` i think. but one could obviously adapt the helpers here.

anyway, i'm really happy about that feature :-)

whatyouhide:
@manukall so out of curiosity, what would be a good return value of `(d)ngettext_noop` in your use case?

manukall:
i'm not really sure.

ecto just adds one string that works for both singular and plural (`"should be %{count} character(s)"`) and the `ErrorHelpers` pass that to `dngettext` twice:

`Gettext.dngettext(ExtatServer.Gettext, "errors", msg, msg, count, opts)`

i don't think returning only the singular or plural string from `dngettext_noop` makes any sense, though. so i can't think of any better solution than your's plus adjusting the `ErrorHelpers`.

whatyouhide:
Awesome @manukall, let us know how it goes with this implementation then :)

PR #144, opened 2017-01-09
Support of compiler_po_wildcard option in Gettext.Compiler

nsi88:
Useful to spend less RAM in dev during compilation
The code looks a litlle bit tricky because of different po_wildcard format in Mix.Tasks.Compile.Gettext and Gettext.Compiler

### DIFF
--- a/lib/gettext/compiler.ex
+++ b/lib/gettext/compiler.ex
@@ -29,6 +29,14 @@ defmodule Gettext.Compiler do
     opts = Keyword.merge(mix_config_opts, compile_time_opts)
 
     priv = Keyword.get(opts, :priv, @default_priv)
+    wildcard = Keyword.get(opts, :compiler_po_wildcard)
+    # NOTE: compiler_po_wildcard format: "gettext/*/LC_MESSAGES/*.po"
+    if wildcard do
+      [_ | wildcard] = Path.split(wildcard)
+      wildcard = Path.join(wildcard)
+    else
+      wildcard = @po_wildcard
+    end
     plural_forms = Keyword.get(opts, :plural_forms, Gettext.Plural)
 
     translations_dir = Application.app_dir(otp_app, priv)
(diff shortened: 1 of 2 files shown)


lib/gettext/compiler.ex
```diff
@@ -29,6 +29,14 @@ defmodule Gettext.Compiler do
     opts = Keyword.merge(mix_config_opts, compile_time_opts)
 
     priv = Keyword.get(opts, :priv, @default_priv)
+    wildcard = Keyword.get(opts, :compiler_po_wildcard)
+    # NOTE: compiler_po_wildcard format: "gettext/*/LC_MESSAGES/*.po"
+    if wildcard do
+      [_ | wildcard] = Path.split(wildcard)
```
josevalim: I don't think we should silently ignore the first directory. We should document that the compiler_po_wildcard is used from inside `priv`.

josevalim: Oh, I see why you did it. It is because we already have a compiler_po_wildcard option.

josevalim: Btw, can you tell us which value you are using for this option?

whatyouhide:
I wonder: does this PR show the need for a unified wildcard (or list of wildcards) where Gettext should look for PO files? Maybe it shouldn't be both per-backend for the POs that we compile into BEAM and also configured for the Mix compiler 🤔

lib/gettext/compiler.ex
nsi88: I use "gettext/ru/LC_MESSAGES/*.po" for dev
Actually we have 18 locales with about 600 keys each. So gettext eats upto 5 GB of RAM during compilation

josevalim:
@nsi88 sorry for the delay on this. Can you please clarify why the wildcard allows you to consume less memory?

nsi88:
@josevalim It's ok. I had enough time to think about it) The wildcard allows to limit the number of locales in dev env, making development faster.

josevalim:
@nsi88 if that's the goal, I would rather have an option that filter locales. Does that make sense?

nsi88:
@josevalim Yeah, it's a good idea. Thank you

josevalim:
@nsi88 can you plaese send a new PR then? :)

nsi88:
@josevalim I will try to find time

josevalim:
Closing this for now. If it is still an interest, please send a new PR. Thanks!

PR #145, opened 2017-01-09
Test on Elixir 1.4.0

xtian:
I also updated the .travis.yml to match elixir-lang/ex_doc#656 so that only the newest and oldest Elixir/OTP pairs are tested.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,8 @@ matrix:
   include:
     - otp_release: 18.3
       elixir: 1.2.6
-    - otp_release: 18.3
-      elixir: 1.3.4
-    - otp_release: 19.1
-      elixir: 1.3.4
+    - otp_release: 19.2
+      elixir: 1.4.0
 
 sudo:
   false


whatyouhide:
Personally I have a preference for including a specific build matrix with explicit versions instead of defaulting and then adding more versions. Is there any benefit in doing what this PR does?

josevalim:
I prefer to list two versions and always do implicit inclusions in the matrix. Otherwise it is very easy to end-up with a very large and unnecessary build matrix.

Someone needs to think about the planet.

xtian:
@whatyouhide No benefit, I just made it match a few of the other elixir-lang projects. I switched it back to the previous formatting.

whatyouhide:
@xtian since the idea is to test the earliest supported version and the latest supported version, can we take advantage of this PR and update Erlang 19.1 to Erlang 19.2 as well? :)

xtian:
@whatyouhide Updated!

whatyouhide:
Thanks a lot @xtian! 💟

PR #148, opened 2017-02-01
Introduce a global locale (per-process) for all Gettext backends

whatyouhide:
This commit introduces the idea of a "global" Gettext locale that spans across all backends. I introduced `Gettext.put_locale/1`, `Gettext.get_locale/0`, and `Gettext.with_locale/2` to interact with it. Now, the algorithm to get a locale when calling Gettext functions for a given backend is roughly this:

    pdict_for_backend || pdict_global || default_for_backend || default_global

A few notable things:

  * I removed the optimization in `Gettext.get_locale/1` that stored the default locale in the pdict in case there was no pdict; this conflicted with the algorithm above. Now, every call to `get_locale/1` with no previous call to `put_locale/1,2` will result in one or two ETS fetches (from the app configs).

  * I changed the behaviour of `with_locale/2,3` slightly to adjust to the change above: now we don't always put the locale back after executing the given function, but we do so only if the locale was there in the first place. This means I had to stop using `get_locale/0,1` directly in with_locale because those functions always return a locale, so there's no information on whether the locale was set in the pdict or was taken from the default one.

### DIFF
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -83,23 +83,39 @@ defmodule Gettext do
   ## Locale
 
   At runtime, all gettext-related functions and macros that do not explicitely
-  take a locale as an argument read the locale from `Gettext.get_locale/1`. The
-  locale can be set with `Gettext.put_locale/2`. Locales are expressed as
-  strings (like `"en"` or `"fr"`); they can be arbitrary strings as long as they
-  match a directory name.
+  take a locale as an argument read the locale from `Gettext.get_locale/0` and
+  `Gettext.get_locale/1`. The locale can be set with `Gettext.put_locale/1` and
+  `Gettext.put_locale/2`. Locales are expressed as strings (like `"en"` or
+  `"fr"`); they can be arbitrary strings as long as they match a directory name.
 
-  Gettext stores the locale **per-process** (in the process dictionary) and per
-  Gettext module. This means that `Gettext.put_locale/2` must be called in every
-  new process in order to have the right locale available in that process. Pay
-  attention to this behaviour, since not setting the locale with
-  `Gettext.put_locale/2` *will not* result in any errors when `Gettext.get_locale/1`
-  is called; the default locale will be returned instead.
+  Gettext stores the locale **per-process** (in the process dictionary). It
+  stores both a global (still per-process) locale that all Gettext backends use
+  by default, as well as a backend-specific locale.  Since the locale is stored
+  per-process, `Gettext.put_locale/1` or `Gettext.put_locale/2` must be called
+  in every new process in order to have the right locale available in that
+  process. Pay attention to this behaviour, since not setting the locale *will
+  not* result in any errors when `Gettext.get_locale/0` or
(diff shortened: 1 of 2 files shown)


lib/gettext.ex
```diff
@@ -83,31 +83,56 @@ defmodule Gettext do
   ## Locale
 
   At runtime, all gettext-related functions and macros that do not explicitely
```
josevalim: The docs look good but I think we need to give focus on the priority upfront. My suggestion:

> At runtime, all gettext-related functions and macros that do not explicitly
> take a locale as an argument read the locale from the backend local and
> fallbacks to Gettext's locale.

> The function `Gettext.put_locale/1` can be used to change the locale of
> all backends for the current Elixir process. That's the preferred mechanism
> for setting the locale at runtime. `Gettext.put_locale/2` can be used when
> you want to change the locale of one specific backend without affecting
> others.

> Similarly, `Gettext.get_locale/0` gets the locale for all backends in the current
> process. `Gettext.get_locale/1` gets the locale of a specific backend for the
> current process. Check their documentation for more information.

lib/gettext.ex
```diff
@@ -472,6 +497,53 @@ defmodule Gettext do
...
+  @spec put_locale(locale) :: nil
+  def put_locale(locale) when is_binary(locale),
+    do: Process.put(Gettext, locale)
+  def put_locale(locale),
+    do: raise(ArgumentError, "put_locale/1 only accepts binary locales, got: #{inspect locale}")
+
+  @doc """
   Gets the locale for the current process and the given backend.
 
   This function returns the value of the locale for the current process and the
```
josevalim: This documentation is out of date. Please check the whole paragraph.

lib/gettext.ex
```diff
@@ -472,6 +497,53 @@ defmodule Gettext do
...
+      #=> "en"
+
+  """
+  @spec get_locale() :: locale
+  def get_locale() do
+    if locale = Process.get(Gettext) do
+      locale
+    else
+      # If this is not set by the user, it's still set in mix.exs (to "en").
+      Application.get_env(:gettext, :default_locale)
```
josevalim: Use `Application.fetch_env!` then or can we have bootstrap issues during compilation?

lib/gettext.ex
```diff
@@ -489,16 +561,24 @@ defmodule Gettext do
...
+    cond do
+      locale = Process.get(backend) ->
+        locale
+      global_locale = Process.get(Gettext) ->
+        global_locale
+      default_locale = get_default_backend_locale(backend) ->
+        default_locale
+      true ->
+        # If this is not set by the user, it's still set in mix.exs (to "en").
+        Application.get_env(:gettext, :default_locale)
```
josevalim: Use `Application.fetch_env!` then or can we have bootstrap issues during compilation?

lib/gettext.ex
whatyouhide: I had just took `fetch_env!` out of the working area of my brain. Tested it on an app using Gettext and on a freshly compiled Gettext as well and it appears to work. :)

PR #150, opened 2017-02-26
Warn when compiling and raise at runtime for missing plural forms

whatyouhide:
If a translation for a given locale now is missing one or more plural forms for that locale, we will warn when compiling the Gettext backend and raise at runtime if the user tries to use the missing plural form.

### DIFF
--- a/lib/gettext/compiler.ex
+++ b/lib/gettext/compiler.ex
@@ -29,7 +29,7 @@ defmodule Gettext.Compiler do
     opts = Keyword.merge(mix_config_opts, compile_time_opts)
 
     priv = Keyword.get(opts, :priv, @default_priv)
-    plural_forms = Keyword.get(opts, :plural_forms, Gettext.Plural)
+    plural_mod = Keyword.get(opts, :plural_forms, Gettext.Plural)
 
     translations_dir = Application.app_dir(otp_app, priv)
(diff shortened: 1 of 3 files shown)


lib/gettext/compiler.ex
```diff
@@ -294,19 +293,35 @@ defmodule Gettext.Compiler do
...
         end
       end
     end
   end
 
+  defp warn_if_missing_plural_forms(locale, plural_mod, translation, file) do
+    Enum.each(0..plural_mod.nplurals(locale) - 1, fn form ->
+      unless Map.has_key?(translation.msgstr, form) do
+        Logger.error([
+          "#{file}:#{translation.po_source_line}: translation is missing plural form ",
```
josevalim: No need for interpolation if you are using an io list? Also, maybe `Logger.warn`?

whatyouhide: @josevalim I interpolated for not doing the `file, ?:, Integer.to_string(line), ": ..."` dance, but yeah we could do that. I used `Logger.error` instead of `warn` because of https://github.com/elixir-lang/gettext/blob/292d020444115239f75f167111de28f725cd2ea6/lib/gettext/compiler.ex#L236; I actually think we should not use logger but just `IO.warn` (and possibly bump to Elixir ~> 1.3 then), but I remember we discussed this and went with Logger but can't remember why. :|

PR #152, opened 2017-03-08
fix `put_env` arity

chenxsan:
check https://hexdocs.pm/elixir/Application.html#put_env/4


whatyouhide:
Technically they both point to the same documentation, but we tend to use the highest arity of a function with default arguments, so good catch :). Thanks a lot! 💟

PR #154, opened 2017-03-25
flags can be separated by [, ]+

roques:
GNU `msgmerge` separates flags by `", "`, do this on output as well,
but accept `[, ]+` as a separator on input.

When parsing multiple flags of a file generated by `msgmerge`, the separating `","` is currently parsed as being part of the preceding flag, which is obviously not what was intended by `msgmerge`.

### DIFF
--- a/lib/gettext/po.ex
+++ b/lib/gettext/po.ex
@@ -280,8 +280,8 @@ defmodule Gettext.PO do
       flags =
         flags
         |> Enum.sort
-        |> Enum.map(&[?\s, &1])
-      ["#,", flags, ?\n]
+        |> Enum.intersperse(", ")
+      ["#, ", flags, ?\n]
     end
   end
 
--- a/lib/gettext/po/parser.ex
+++ b/lib/gettext/po/parser.ex
@@ -75,7 +75,7 @@ defmodule Gettext.PO.Parser do
   defp parse_flags(flag_comments) do
     flag_comments
     |> Stream.map(fn("#," <> content) -> content end)
-    |> Stream.flat_map(&String.split(&1, ~r/\s+/, trim: true))
+    |> Stream.flat_map(&String.split(&1, ~r/[,\s]+/, trim: true))
     |> Enum.into(MapSet.new)
   end
 
--- a/test/gettext/po/parser_test.exs
+++ b/test/gettext/po/parser_test.exs
@@ -178,15 +178,15 @@ defmodule Gettext.PO.ParserTest do
 
   test "flags are extracted in to the :flags field of a translation" do
     parsed = Parser.parse([
-      {:comment, 1, "#, flag other-flag"},
+      {:comment, 1, "#, flag,a-flag b-flag, c-flag"},
       {:comment, 2, "# comment"},
-      {:comment, 3, "#, other-flag other-other-flag"},
+      {:comment, 3, "#, flag,  ,d-flag ,, e-flag"},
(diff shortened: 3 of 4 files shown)


lib/gettext/po.ex
```diff
@@ -279,8 +279,8 @@ defmodule Gettext.PO do
       flags =
         flags
         |> Enum.sort
-        |> Enum.map(&[?\s, &1])
-      ["#,", flags, ?\n]
```
whatyouhide: Do we need to change how we dump this? As far as I understand `msgmerge` accepts space-separated flags, right?

lib/gettext/po/parser.ex
```diff
@@ -75,7 +75,7 @@ defmodule Gettext.PO.Parser do
   defp parse_flags(flag_comments) do
     flag_comments
     |> Stream.map(fn("#," <> content) -> content end)
-    |> Stream.flat_map(&String.split(&1, ~r/\s+/, trim: true))
+    |> Stream.flat_map(&String.split(&1, ~r/[,\s]+/, trim: true))
```
whatyouhide: With this regex, we allow any number of commas/spaces.

```elixir
String.split("foo, , bar,,baz", ~r/[,\s]+/, trim: true)
#=> ["foo", "bar", "baz"]
```

I think a stronger approach would be to split on something like `(\s+,\s*)|(\s*,\s+)`, right?

roques: I just checked, `masmerge` accepts any combination of comma and space, even things like
```pot
#, fuzzy,,, , ,no-c-format     no-python-format,,,,,no-elisp-format
msgid "foo"
msgstr ""
```
… and canonicalizes them to be separated by `", "` on output.
Also, imagine we encounter flags like `foo, ,bar` and our split would not consume the whole of `", ,"`, then the leftover would be part of one of the flags, which very likely is not what the author/editor of the `.pot` intended.

lib/gettext/po.ex
roques: Yes, `msgmerge` accepts `" "` as separator when parsing.  But, as it always uses `", "` as separator on output, I can imagine that other `.po`-parsing tools might not be as forgiving and expect `", "`. Behaving like `msgmerge` should be the safest way.

whatyouhide:
@roques have you tried to see what tools like POEdit or Transifex do and how they format flags? I don't have the time to do it this week as I am traveling, if you haven't, can you check their behaviour? 💟

roques:
@whatyouhide I just tested with Poedit: When reading, it accepts and understands all kinds of combinations of commas and spaces as separators for flags. When writing it canonicalizes them to `", "`.

lib/gettext/po.ex
```diff
@@ -279,8 +279,8 @@ defmodule Gettext.PO do
...
+        |> Enum.map(&[?,, ?\s, &1])
+      ["#", flags, ?\n]	# the , of #, is dumped with the first flag
```
whatyouhide: This code is a bit tricky to understand IMO, with the `,` not attached to `#` hiding a bit of the format of PO(T) files. Let's rewrite it as such:

```elixir
flags =
  flags
  |> Enum.sort()
  |> Enum.intersperse(", ")
["#, ", flags, ?\n]
```

roques: Yes, that's much easier to understand.

whatyouhide:
Thanks @roques! 💟

PR #156, opened 2017-03-25
[WIP] Extracted comments

roques:
The GNU gettext suite can extract some specially marked comments from the program's source to be included into `.pot` and `.po` files as comments *to* the translators. This has proven to be very valuable in past projects.
This branch currently contains two commits, one introducing handling, parsing, and dumping of extracted comments from/to `.po` files and a second commit implementing the extraction itself. I don't really like the way I've implemented the actual extraction process and would like some feedback how to improve it.
Currently, source like this:
```ex
def cocoballs(n) do
  gettext "Cocoballs are delicious!",
    %{comment: "These are Guyanese Cocoa balls, made of raw cocoa to be used when making *real* hot chocolate."}
```
will result in a `.pot` like:
```po
#. These are Guyanese Cocoa balls, made of raw cocoa to be used when making *real* hot chocolate.
msgid "Cocoballs are delicious!"
msgstr ""
```
Do not confuse extracted comments with comments *from* translators, which are marked like:
```po
# How big are these? Depending on size they could be "Bälle" or "Kugeln"
msgid "Cocoballs are delicious!"
msgstr "Kakaokugeln sind lecker!"
```

### DIFF
--- a/lib/gettext/compiler.ex
+++ b/lib/gettext/compiler.ex
@@ -60,44 +60,46 @@ defmodule Gettext.Compiler do
 
   defp macros do
     quote unquote: false do
-      defmacro dgettext_noop(domain, msgid) do
+      defmacro dgettext_noop(domain, msgid, bindings \\ Macro.escape(%{})) do
         domain = Gettext.Compiler.expand_to_binary(domain, "domain", __MODULE__, __CALLER__)
         msgid = Gettext.Compiler.expand_to_binary(msgid, "msgid", __MODULE__, __CALLER__)
+        bindings = Gettext.Compiler.expand_bindings(bindings, __CALLER__)
 
         if Gettext.Extractor.extracting? do
-          Gettext.Extractor.extract(__CALLER__, __MODULE__, domain, msgid)
+          Gettext.Extractor.extract(__CALLER__, __MODULE__, domain, msgid, bindings)
         end
 
         msgid
       end
 
-      defmacro gettext_noop(msgid) do
+      defmacro gettext_noop(msgid, bindings \\ Macro.escape(%{})) do
         quote do
-          unquote(__MODULE__).dgettext_noop("default", unquote(msgid))
+          unquote(__MODULE__).dgettext_noop("default", unquote(msgid), unquote(bindings))
         end
       end
 
-      defmacro dngettext_noop(domain, msgid, msgid_plural) do
+      defmacro dngettext_noop(domain, msgid, msgid_plural, bindings \\ Macro.escape(%{})) do
         domain = Gettext.Compiler.expand_to_binary(domain, "domain", __MODULE__, __CALLER__)
         msgid = Gettext.Compiler.expand_to_binary(msgid, "msgid", __MODULE__, __CALLER__)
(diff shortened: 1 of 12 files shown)


josevalim:
Thanks @roques!

I am actually not a fan of repurposing bindings for adding comments. It would need to be a third argument. But between repurposing bindings, a third argument and adding a gettext_comment macro, I would probably use a `gettext_comment` macro.

For now, I think adding the comments to the POT is likely the way to go.

roques:
I'm not sure I completely understand how you envision `gettext_comment()` to work. Which one of the following do you mean:
 1. `gettext_comment()` only adds a comment to an entry to the POT:
    ```elixir
    gettext_comment("cocoballs", "The Caribbean way of making hot chocolate")
    gettext("coco-balls")
    ```
 2. `gettext_comment()` behaves like `gettext()`, but also sets a POT comment:
    ```elixir
    gettext_comment("cocoballs", "The Caribbean way of making hot chocolate")
    ```
 3. `gettext_comment()` is used to annotate the string to be translated:
    ```elixir
    "cocoballs"
    |> gettext_comment("The Caribbean way of making hot chocolate")
    |> gettext()
    ```
Of these I would prefer 2, as 1 requires msg_id to be repeated, and 3 separates it too much from the `gettext()`.  However, 2 is just an unmagical version of looking at the type of the 2nd argument to determine if it's bindings or a comment: If we would do that, the signature of `gettext()` could be:
```elixir
gettext(msg_id, bindings \\ %{}, comment \\ nil)
```
This is a bit more magical, but I'm not sure if it would be too confusing. It does open the door for mistakes like the following however:
```elixir
gettext("Hello, %{name}", "Herbert")
```
when what was actually meant was:
```elixir
gettext("Hello, %{name}", %{name: "Herbert"})
```

Some more guidance would be appreciated.

josevalim:
The problem with 2 is that it would require also a version for dgettext, dngettext and so on.

Passing it as third argument is also OK in my opinion. But we definitely should not mix runtime bindings and compile-time options.

whatyouhide:
Just a quick note: passing it as an additional argument to current Gettext macros is something I would really like to avoid since the macros today are already quite complex-arity wise and have default arguments. I like the idea of decorating a string with a comment like

```elixir
gettext(gettext_comment("my msgid", "some comment"), ...)
```

where `gettext_comment/2` would return something like `{:commented_string, "my msgid", "some comment"}` and the Gettext macros would be updated to take that into consideration.

josevalim:
@whatyouhide one other approach that would support both 1 and 3 is to make gettext_comment store the comment somewhere and make it return the msgid directly. I am not sure how much of a fan I am of having a special tuple.

whatyouhide:
@josevalim I thought about that too. I think we could store the comment in the pdict since the compiler will expand the two macros in the same process. Were you thinking about this as well? Or maybe storing the comment inside say an ETS/Agent/whatever under the msgid key, and reading from Gettext macros?

josevalim:
@whatyouhide pdict sounds good!

whatyouhide:
Closing in favor of https://github.com/elixir-lang/gettext/pull/158, whose work is heavily based on this PR. Thanks a lot @roques! 💟

PR #158, opened 2017-06-13
Add support for extracted comments via gettext_comment/1

whatyouhide:
Based on the great work by @roques in #156. 💟

### DIFF
--- a/lib/gettext/po/plural_translation.ex
+++ b/lib/gettext/po/plural_translation.ex
@@ -19,6 +19,8 @@ defmodule Gettext.PO.PluralTranslation do
       `Gettext.Plural`.
     * `comments` - a list of comments as they are found in the PO file (e.g.,
       `["# foo"]`).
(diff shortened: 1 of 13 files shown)


lib/gettext/po/plural_translation.ex
```diff
@@ -19,6 +19,8 @@ defmodule Gettext.PO.PluralTranslation do
       `Gettext.Plural`.
     * `comments` - a list of comments as they are found in the PO file (e.g.,
       `["# foo"]`).
+    * `extracted_comments` - a list of extracted comments (for example,
+      `["foo", "bar"]`).
```
roques: I think the code stores these as `["#. foo", "#. bar"]`.

roques:
@whatyouhide Grazie mille for finishing this. I like this better than my prototype.

lib/gettext/po/plural_translation.ex
whatyouhide: Great catch, updated. :)

Issue #159, opened 2017-08-31
Consider making comments in .po files configurable / optional


manukall:
The current behavior of adding all the occurrences as `filename:linenum` comments to the .po files leads to  huge commits every time we add just a single new translation in our app. 

Usually we've also changed our app in different places in the meantime and all the line numbers are different now, so basically every comment in every .po file is changed when runnning `mix gettext.extract --merge`.

I'd love for a way to either make the comments completely optional or configurable in a way that lets us turn off line numbers.

whatyouhide:
We use `file:line` because that's what GNU gettext uses. Have you researched if tools (like POEdit) understand reference comments then the format for such comments is not `file:line`?

manukall:
I have not as I wasn't aware that was an official thing. After reading https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html, 

I think maybe it makes more sense to just allow turning those reference comments off. Above document states "All comments, of either kind, are optional".

whatyouhide:
Making reference comments optional might be a viable way of solving the problem, but it does introduce a bit of complexity in how Gettext works. @josevalim thoughts?

josevalim:
Sounds good, yes.

whatyouhide:
@manukall are you willing to work on a PR for this? :)

manukall:
@whatyouhide i've got a pretty full schedule for the next weeks, so anything complicated during that time isn't realistic. if that's a smaller change or if it has time until later, i'd absolutely be willing, though.

do you have an idea about the implementation, yet? i guess simply checking `Application.get_config(...)` in `Gettext.PO.dump_references/1` wouldn't be good enough, right? ;-)

whatyouhide:
It wouldn't yeah. We moved all Gettext configuration to `mix.exs` under the `:gettext` key so we would need to store this information there and pass that around to the right places. I can work on it, no worries.

whatyouhide:
Closing in favour of #161.

PR #161, opened 2017-10-06
Make reference comments optional

whatyouhide:
Related to #159.


whatyouhide:
The failure on Elixir 1.2 is fine as I plan to drop support for Elixir < 1.3 right after merging this PR.

padde:
@whatyouhide I am currently experimenting with the `write_reference_comments: false` setting to avoid very time consuming merge conflicts in PO files. It seems that `mix gettext.merge` will unfortunately discard existing `fuzzy` and `elixir-format` tags. Is this behavior intended?

PR #164, opened 2017-11-03
Define intermediate functions per locale + domain and support parallel compilation

josevalim:
This speeds up compilation time about roughly 20% in large gettext backends.

On the sample project we had, the original time compilation time of the backend was 30s, which falls down to 24s with the first commit.

By using the option in the second commit, compilation time drops to 11s (my machine has 2 cores, the more than 2x increase is probably thanks to the smaller beam files).

### DIFF
--- a/lib/gettext/compiler.ex
+++ b/lib/gettext/compiler.ex
@@ -55,9 +55,11 @@ defmodule Gettext.Compiler do
       end
 
       unquote(macros())
+      unquote(catch_all_helpers())
+
       unquote(signatures())
       unquote(compile_po_files(translations_dir, plural_mod))
-      unquote(dynamic_clauses())
+      unquote(catch_all_clauses())
     end
   end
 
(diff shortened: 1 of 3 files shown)


lib/gettext/compiler.ex
```diff
@@ -247,17 +263,92 @@ defmodule Gettext.Compiler do
...
+      |> Enum.each(&Task.await(&1, :infinity))
+
+      quoted
+    else
+      Enum.map(po_files, &compile_serial_po_file(&1, plural_mod))
+    end
+  end
+
+  defp create_locale_module(env, {module, translations}) do
+    Module.create(module, block([catch_all_helpers() | translations]), env)
```
whatyouhide: Do we need `@moduledoc false`?

josevalim: Good catch, yes.

PR #167, opened 2017-11-26
Add option domain_prefix for backends

hrom512:
Added option `domain_prefix` for backends, to remove duplication in `dgettext`.

Before:

```elixir
defmodule Admin.Gettext do
  use Gettext, otp_app: :my_app
end

Admin.Gettext.dgettext("admin.default", "Page title")

Admin.Gettext.dgettext("admin.articles", "Title")

Admin.Gettext.dgettext("admin.errors", "Email is invalid")
```

After:

```elixir
defmodule Admin.Gettext do
  use Gettext, otp_app: :my_app, domain_prefix: "admin."
end

Admin.Gettext.gettext("Page title")

Admin.Gettext.dgettext("articles", "Title")

Admin.Gettext.dgettext("errors", "Email is invalid")
```

### DIFF
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -505,12 +505,16 @@ defmodule Gettext do
   @type backend :: module
   @type bindings :: %{} | Keyword.t
 
+  @default_domain "default"
+
   @doc false
   defmacro __using__(opts) do
     quote do
       require Logger
 
       @gettext_opts unquote(opts)
+      @default_domain "default"
+
       @before_compile Gettext.Compiler
 
       def handle_missing_bindings(exception, incomplete) do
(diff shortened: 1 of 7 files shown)


whatyouhide:
I'm not a fan of this feature as it adds magic to the behaviour of a Gettext backend. Maybe we can make the `*gettext` macros overridable but I'm not sure that's a good idea either. Thoughts @josevalim?

hrom512:
I'm new to elixir and so I'm not sure how to properly solve such problems in elixir. I will be grateful for any help.

whatyouhide:
I am not in favour of adding this feature as I said since it introduces magic that I find not worth introducing. I have no better solution other than writing the full domain, which makes things clearer and less abstract. I'll go ahead and close this but if other folks have better ideas on how to solve this or share this need, we can reopen the discussion. Thanks @hrom512! 💟

PR #174, opened 2018-02-15
Generate correct plural forms when dumping new translations in PO files

### LINKED ISSUE #151: Generation of right plural forms for locale
`Gettext` generates .po-files for Russian locale with insufficient plural forms (Russian needs three, but only two generated).

whatyouhide:
Closes #151.


lib/mix/tasks/gettext.merge.ex
```diff
@@ -125,9 +134,9 @@ defmodule Mix.Tasks.Gettext.Merge do
     if Path.extname(arg1) == ".po" and Path.extname(arg2) in [".po", ".pot"] do
```
josevalim: Should we bind arg1 and arg2 to variables likes source and target?

lib/gettext/merger.ex
```diff
@@ -166,34 +181,54 @@ defmodule Gettext.Merger do
...
+    Keyword.put_new_lazy(opts, :plural_forms, fn ->
+      read_plural_forms_from_headers(headers) || Gettext.Plural.nplurals(locale)
+    end)
+  end
+
+  defp read_plural_forms_from_headers(headers) do
+    Enum.find_value(headers, fn header ->
+      with "Plural-Forms:" <> rest <- header,
+           "nplurals=" <> rest <- String.trim(rest),
+           {plural_forms, _rest} <- Integer.parse(rest) do
```
josevalim: 😍

lib/mix/tasks/gettext.merge.ex
whatyouhide: Yes definitely, this sucks. I was keeping the renaming for a bunch of refactorings I'm gonna do after merging this. 👍

PR #181, opened 2018-07-07
Implement tag bindings

LostKobrakai:
This allows bindings to be 1-arity functions receiving the translated text of `%{binding:translatable content}`. This would make it possible to have e.g. links being assembled around some word within a sentence without the need to break the sentence up into multiple parts, which would need to be individually assembled or directly embedding html in the translatable content.

```
# Tag bindings 
msgid = "I lived in %{wiki:Italy} for a year"
wiki = &(~s[<a href="https://en.wikipedia.org/wiki/Italy">#{&1}<a>])

assert Translator.lgettext("it", "interpolations", msgid, %{wiki: wiki}) ==
         {:ok, ~s[Ho vissuto in <a href="https://en.wikipedia.org/wiki/Italy">italia<a> per un anno]}
```

Things still to be done:

* [ ] Handling the new error case in `gettext.ex`
* [ ] Update docs

### DIFF
--- a/lib/gettext/compiler.ex
+++ b/lib/gettext/compiler.ex
@@ -527,6 +527,9 @@ defmodule Gettext.Compiler do
       key, acc when is_atom(key) ->
         quote do: unquote(acc) <> to_string(unquote(Macro.var(key, __MODULE__)))
 
+      {key, inner}, acc when is_atom(key) ->
+        quote do: unquote(acc) <> to_string(unquote(Macro.var(key, __MODULE__)).(unquote(inner)))
+
       str, acc ->
         quote do: unquote(acc) <> unquote(str)
     end)
--- a/lib/gettext/interpolation.ex
+++ b/lib/gettext/interpolation.ex
@@ -22,16 +22,19 @@ defmodule Gettext.Interpolation do
   """
   @spec to_interpolatable(String.t()) :: interpolatable
   def to_interpolatable(string) do
-    start_pattern = :binary.compile_pattern("%{")
-    end_pattern = :binary.compile_pattern("}")
+    patterns = %{
+      start: :binary.compile_pattern("%{"),
+      end: :binary.compile_pattern("}"),
+      nested_delimiter: :binary.compile_pattern(":")
+    }
 
     string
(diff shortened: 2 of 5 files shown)


whatyouhide:
I understand the use case behind this feature, but I feel the API would be very confusing like this. @josevalim thoughts?

josevalim:
Agreed. It does give the impression that this is backwards. Wouldn't it be preferred for the gettext to return the annotated content that you then replace later?

I.e. make the translation return `%{wiki:italia}` and then you replace it by the link.

LostKobrakai:
The translation in the .po file is `%{wiki:italia}`. The link is put around the translation by gettext.

josevalim:
@LostKobrakai sorry for not being clear. What I am saying is that you can do another pass later and replace the returned `%{wiki:italia}` contents. There isn't a strong reason to make this part of gettext per se.

josevalim:
We could open up a case to make this part of gettext but, if so, we would like to see how other gettext implementations handle this.

josevalim:
@LostKobrakai right. It can be done comfortably outside of gettext. Probably in fewer lines of code than what we have right now. So my vote it indeed to due it as a separate pass. :)

whatyouhide:
I agree with @josevalim, seems like having this feature in Gettext wouldn't bring great benefits. Also, if we can get away with **not** adding a feature, it means less complexity :). Thanks for your work @LostKobrakai! 💟

LostKobrakai:
@josevalim do you have any hints on how to tackle that from outside of gettext? At best I'd like to still implement `Gettext.Backend` so the user can just swap aliases / imports.

josevalim:
You can do: `gettext("...") |> replace(...)` If that's too much, you can also create your own rgettext or similar that you import.

LostKobrakai:
I think just doing the replace after gettext is done will leave the user with lot's of logged errors for missing bindings.

josevalim:
@LostKobrakai can't you use a syntax that is not the gettext one?

LostKobrakai:
Yeah, realized that a few minutes ago :D

PR #185, opened 2018-08-08
Fix bindings type.

xadhoom:
%{} is the type for an empty map, we should use map() instead.


xadhoom:
Discovered while using maps instead of kw in my `gettext/2` calls.

josevalim:

PR #186, opened 2018-08-20
Support filenames with colons in reference comments

### LINKED ISSUE #184: Warning when there's a : in a filename
I just noticed `:` in filenames break gettext. I'm not sure if that can be supported, but at least gettext should issue a warning if it sees such a file.

whatyouhide:
Closes #184.

### DIFF
--- a/lib/gettext/po.ex
+++ b/lib/gettext/po.ex
@@ -257,15 +257,11 @@ defmodule Gettext.PO do
   end
 
   defp dump_references(references, gettext_config) do
-    if Keyword.get(gettext_config, :write_reference_comments, true) do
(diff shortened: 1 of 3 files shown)


lib/gettext/po.ex
```diff
@@ -257,17 +257,13 @@ defmodule Gettext.PO do
   end
 
   defp dump_references(references, gettext_config) do
-    if Keyword.get(gettext_config, :write_reference_comments, true) do
+    if Keyword.get(gettext_config, :write_reference_comments, true) and references != [] do
       dump_references(references)
     else
-      dump_references([])
+      ""
```
whatyouhide: This is just a refactoring.

PR #187, opened 2018-08-21
Add handle_missing_translation/4,6 callbacks to Gettext backends

whatyouhide:
I went with two callbacks, `handle_missing_translation/4` for translations and `handle_missing_translation/6` for plural translations. I don't think that the approach with the exception, similar to `handle_missing_bindings/2`, worked very well in this case. We don't warn or don't produce an error message by default when a translation is missing, so the exception would feel a bit forced.

@josevalim a point one could make is that a user might want to actually raise on every missing translation, and in that case having a single exception (in a single callback, not two) that they can just raise might be beneficial. If you still think that's the better approach, it'll be an easy change to this PR :).

### DIFF
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -528,6 +528,28 @@ defmodule Gettext do
       end
 
       defoverridable handle_missing_bindings: 2
+
+      def handle_missing_translation(_locale, domain, msgid, bindings) do
+        import Gettext.Interpolation, only: [to_interpolatable: 1, interpolate: 2]
+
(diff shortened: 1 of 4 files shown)


josevalim:
Should we call it then `handle_missing_msgid` and `handle_missing_plural_msgid` or similar?

whatyouhide:
@josevalim good idea on the different names. Maybe `handle_missing_translation` and `handle_missing_plural_translation` then?

josevalim:
Sounds good to me. :+1:

whatyouhide:
@josevalim updated.

PR #195, opened 2018-12-13
Report changes on merge

LostKobrakai:
#193

PR #201, opened 2019-06-15
Add stats reporting when merging

### LINKED ISSUE #193: Report changes on merge
I'm wondering if it would be possible to report changes when merging new translations into .po files. My specific problem is that sometimes merge twice or more in succession without going through all files and if I understand that correctly a merge will mark as fuzzy match translations only once as to be looked at and a second merge would remove that marking. Having something like 

```
Wrote priv/gettext/en/LC_MESSAGES/default.po:
14 new / 4 fuzzy / 5 removed / 988 unchanged
```

Would be a great way to know what happened and if action needs to be taken.

whatyouhide:
Closes #193.

This PR adds stats reporting when merging PO files. 

```
Wrote tmp/gettext.merge/it/LC_MESSAGES/foo.po (0 exact matches, 0 fuzzy matches, 1 new, 0 removed)
```

It's based on @LostKobrakai's work in #195.

@josevalim I don't love a bunch of the complexity that this adds but building a `stats` map as a functional data structure without doing anything weird is the simplest approach I could think of. Let me know what you think :)


lib/gettext/merger.ex
```diff
@@ -49,35 +49,59 @@ defmodule Gettext.Merger do
...
+                {t, {stats_acc, unused}}
+            end
+
+          :error ->
+            stats_acc = update_in(stats_acc.new, &(&1 + 1))
+            {t, {stats_acc, unused}}
+        end
+      end)
+
+    {translations, put_in(stats.removed, map_size(unused))}
```
whatyouhide: @josevalim such simple code, gone 😭

LostKobrakai: Maybe like this: https://github.com/LostKobrakai/gettext/commit/2d132045f9ce6a9c871ea931d5fea06178e781d1? It keeps the simply map, but does do a few more passes over the data.

whatyouhide: The `length(old.translations) - (length(translations) - stats.new)` code seemed to be buggy when dealing with fuzzy translations (sorry, it's been a while, can't remember in more detail). I think I'd rather do it in one pass anyways because while the code is more complex it's very local.

LostKobrakai: I'm not really sure where your code would result in a different value.

`translations` as the result of `merge_translations` should hold all new/exact_match/fuzzy_match translations. Removing the new ones should give the number of translations, which are exact_match or fuzzy_match. Removing that from the number of old translations should result in the number of removed ones.

Your code does the same. Take the old list and on exact_match or fuzzy_match remove the item.  You're just doing it in the reduce.

Or can there be many translations with the same key?

whatyouhide: If there are two fuzzy translations that match one old translations, `length([_]) - (length(2) - 0)` would result in `-1` removed translations, maybe that was it?

LostKobrakai: Yeah didn‘t think of that situation. Then I can only offer a bit of refactoring to lessen duplicate code needed for the stats gathering: https://github.com/LostKobrakai/gettext/commit/bc1511a5efd02efd72dbc6221eda4efa4ae2345b

Issue #199, opened 2019-05-31
Feature Request: Return multiple po_source_line entries for duplicated msgids


omaciel:
While I'm aware of how multiple msgids are currently handled (#22), it seems that this is common usage for GNOME packages:

* The word `"General"` for the [Epiphany](https://gitlab.gnome.org/GNOME/epiphany/blob/master/po/pt_BR.po) application
 
```
#: src/resources/gtk/prefs-dialog.ui:418
msgid "General"
msgstr "Geral"

#: src/resources/gtk/shortcuts-dialog.ui:15
msgctxt "shortcut window"
msgid "General"
msgstr "Geral"
```

* The word `Default` for [gnome-terminal](https://gitlab.gnome.org/GNOME/gnome-terminal/blob/master/po/pt_BR.po)

```
#: ../src/preferences.ui.h:5
msgctxt "theme variant"
msgid "Default"
msgstr "Padrão"

#. Cursor blink mode
#: ../src/preferences.ui.h:19
msgid "Default"
msgstr "Padrão"
```

It seems to me that whenever you have the presence of a `msgctxt` flag, duplicated entries are split into individual entries, which then causes `PO.parse_file` to raise a `"found a duplicate..."` error.

Would it be acceptable to perhaps return duplicated msgids with multiple `po_source_line` lines to handle these types of situations?

Thanks in advance!

josevalim:
Yeah, we should consider the context when detecting duplicates. A PR is
welcome!
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

omaciel:
Thanks for the quick response @josevalim I am just getting started with Elixir, so it may take a little bit for me to be able to contribute, but I shall try for sure :)

whatyouhide:
This is not a teeny tiny change however, because right now we use the `msgid` or `msgid`/`msgid_plural` as the ID of a translation and we don't even store the `msgctxt` anywhere. The change would mean storing the context with translations and using it as part of the translation key as well.

MeikelLP:
Hey, I'm coming from an issue in Accent (see https://github.com/mirego/accent/issues/113).
I'm trying to translate Unity3D's translation files (.po files) with Accent (which uses this lib). Unity3D does use duplicate keys and they have different meaning depending on the context (code reference).

Is there any ETA?

josevalim:
I believe the bulk of the work was already done in the linked PR above, so
[...]

PR #204, opened 2019-06-30
Store the msgctxt value in translations

whatyouhide:
We were ignoring this value but we might as well store it in the translation structs for future use (see #199).


whatyouhide:
Thank you @whatyouhide! 💟 You're the best!

PR #207, opened 2019-08-09
Allow for duplicates with msgctxt

olafura:
Often you need to control your translations so adding msgctxt allows you
to have different translations even if you are using the same msgid.

Here we are making sure that this doesn't trip the duplication check.

This is hopefully the last bit for this feature request at least on the parsing side:
https://github.com/elixir-lang/gettext/issues/199

### DIFF
--- a/lib/gettext/po/translations.ex
+++ b/lib/gettext/po/translations.ex
@@ -89,27 +89,24 @@ defmodule Gettext.PO.Translations do
 
       iex> t = %Gettext.PO.Translation{msgid: "foo"}
       iex> Gettext.PO.Translations.key(t)
-      "foo"
+      {nil, "foo"}
 
       iex> t = %Gettext.PO.PluralTranslation{msgid: "foo", msgid_plural: "foos"}
       iex> Gettext.PO.Translations.key(t)
-      {"foo", "foos"}
+      {nil, {"foo", "foos"}}
 
       iex> t = %Gettext.PO.PluralTranslation{msgctxt: "bar", msgid: "foo", msgid_plural: "foos"}
       iex> Gettext.PO.Translations.key(t)
       {"bar", {"foo", "foos"}}
   """
-  @spec key(PO.translation()) :: binary | {binary, binary} | {binary, {binary, binary}}
-  def key(%Translation{msgid: msgid, msgctxt: nil}), do: IO.iodata_to_binary(msgid)
+  @spec key(PO.translation()) :: {binary | nil, binary | {binary, binary}}
+  def key(%{msgctxt: msgctxt} = translation), do: {IO.iodata_to_binary(msgctxt), id_key(translation)}
 
-  def key(%Translation{msgctxt: msgctxt} = translation),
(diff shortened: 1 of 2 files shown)


lib/gettext/po/translations.ex
```diff
@@ -97,11 +97,19 @@ defmodule Gettext.PO.Translations do
 
   """
   @spec key(PO.translation()) :: binary | {binary, binary}
-  def key(%Translation{msgid: msgid}), do: IO.iodata_to_binary(msgid)
+  def key(%Translation{msgid: msgid, msgctxt: nil}), do: IO.iodata_to_binary(msgid)
 
-  def key(%PluralTranslation{msgid: msgid, msgid_plural: msgid_plural}),
+  def key(%Translation{msgid: msgid, msgctxt: msgctxt}),
+    do: "#{IO.iodata_to_binary(msgctxt)}_#{IO.iodata_to_binary(msgid)}"
```
josevalim: The problem about doing string concatenation is that we can have false positives. Imagine someone uses the context "my" and msgid "string", and there is another message id called "my_string".

I think the best option would be to nest all existing keys under `msgid`:

    def key(%Translation{msgctxt: msgctxt} = translation) do
      {msgctxt, id_key(translation)
    end

and `id_key` is the current key implementation.

Unless gettext explicitly documents that the msgctx is a concatenation of those?

olafura: I wasn't sure about it because of the two tuple limit but I'll just bump that to three tuple in the spec.

olafura: Sorry I should have read your message better ;)

lib/gettext/po/translations.ex
```diff
@@ -95,13 +95,24 @@ defmodule Gettext.PO.Translations do
...
+      iex> Gettext.PO.Translations.key(t)
+      {"bar", "foo", "foos"}
   """
-  @spec key(PO.translation()) :: binary | {binary, binary}
-  def key(%Translation{msgid: msgid}), do: IO.iodata_to_binary(msgid)
+  @spec key(PO.translation()) :: binary | {binary, binary} | {binary, binary, binary}
+  def key(%Translation{msgid: msgid, msgctxt: nil}), do: IO.iodata_to_binary(msgid)
+
+  def key(%Translation{msgid: msgid, msgctxt: msgctxt}),
+    do: {IO.iodata_to_binary(msgctxt), IO.iodata_to_binary(msgid)}
```
josevalim: You can still have a conflict in this case, although high unlikely, because the key here and below can conflict. My suggestion is to use nested tuples: `{msgctxt, current_key(translation)}`. And always include the msgctxt. it is not a problem.

olafura: This should be fixed now, just had to add the nilling of the translation so we wouldn't get endless recursion.

olafura: I almost forgot to update the spec

josevalim:

PR #208, opened 2019-08-13
Add support for po dump of msgctxt data

olafura:


lib/gettext/po.ex
```diff
@@ -235,6 +235,7 @@ defmodule Gettext.PO do
       dump_comments(t.extracted_comments),
       dump_flags(t.flags),
       dump_references(t.references, gettext_config),
+      dump_kw_and_strings("msgctxt", t.msgctxt),
```
whatyouhide: Instead of adding logic to `dump_kw_and_strings/2`, let's keep that function simple and have:

```elixir
dump_msgctxt(t.msgctxt),
...

# And then
defp dump_msgctxt(nil), do: []
defp dump_msgctxt(string), do: dump_kw_and_strings("msgctxt", string)
```

olafura: Fixed

whatyouhide:
Thank you @olafura! 💟

PR #216, opened 2019-10-20
Optimize Gettext.get_locale/0,1 by explicitly matching on pdict return

lexmag:


josevalim:
"Optimize". :P In any case,  thanks for the PR!

josevalim:

lexmag:
Hehe, was not sure about opening the PR, but technically it is an optimization that removes a couple of jumps—might be helpful with very many calls and slower CPUs. :bowtie:

PR #218, opened 2019-10-22
Preserve whitespaces in flags

lexmag:
See https://github.com/elixir-gettext/gettext/commit/bdd7aff1b9a1088af0131b2c30143621b7f1941d#r35590050.

What do you think about this approach?

### DIFF
--- a/lib/gettext/extractor.ex
+++ b/lib/gettext/extractor.ex
@@ -94,8 +94,7 @@ defmodule Gettext.Extractor do
             "the Gettext backend #{inspect(backend)} has the same :priv directory as " <>
               "#{inspect(other_backend)}, which means they will override each other. " <>
               "Please set the :priv option to different directories on use Gettext " <>
-              "inside each backend",
-            []
+              "inside each backend"
           )
 
         %{} ->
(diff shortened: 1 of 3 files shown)


whatyouhide:
Yeah let's go with warning if whitespace is found and autofix like we do today. Thanks @lexmag!

whatyouhide:
Btw @lexmag, waiting for the PR to not be in the draft state before reviewing/merging 🙃

lexmag:
I've tried several more combinations of flags that contain space and can confirm that both Loco (https://localise.biz) and Transifex.com (transifex.com) parse `#, a b, c` to `["a b", "c"]`.
I think we better to do the same and keep whitespaced flags intact.
I wonder what do we actually get from auto-fixing? What are those tools that have originally influenced our choice?
In fact, if we preserve whitespaces we perhaps can skip the warning altogether (maybe add later if needed).

whatyouhide:
@lexmag I see what you mean now. Yeah, if those tools only split on the comma, let's do the same.

lexmag:
@whatyouhide thanks. Updated.

whatyouhide:
Thanks @lexmag! ❤️❤️❤️

PR #219, opened 2019-10-23
Simplify regex that is used when parsing references

lexmag:


josevalim:

lib/gettext/po/parser.ex
```diff
@@ -60,7 +60,7 @@ defmodule Gettext.PO.Parser do
 
   defp parse_references("#:" <> comment) do
     regex = ~r/
-      ((?:\w|\.|:)(?:\w|\s|\.|:|\/)+?)
+      ([\w\.:][\w\s\.:\/]+?)
```
lexmag: I suppose we also should use `*` instead of `+` to allow `<one-letter>:<line>` references, right?

whatyouhide: Wait, this regex doesn't work with the one letter ref? Boy I hate regexes, I find it so hard to parse this in my head 😂

lexmag: It does not. 😁

whatyouhide: @lexmag, you're my only hope. Just fix the regex however you see fit and add a test 😂

lexmag: Done in #221.

PR #221, opened 2019-10-24
Fix PO parser to support one-letter references

lexmag:


whatyouhide:
Thanks @lexmag ❤️❤️❤️

PR #226, opened 2019-10-28
Consider extracted comments when merging templates during extraction

### LINKED ISSUE #222: gettext_comment disappearing
When using gettext_comment it only appears the first time you extract a msg. 
So for example:

```
    gettext_comment("very important clarifying comment")
    gettext("This is some very awesome text"
```
If you would run gettext.extract the comment does appear in the pot file.
However, if you run extract again the comment is gone.
Only if the string within the gettext call gets updated, and is thus a new msg again, the comment is back in the pot file.

lexmag:
Fixes #222.


whatyouhide:
Thanks @lexmag! 💟

PR #227, opened 2019-11-05
Human copyable format

dmitrydprog:
The previous version of formatting causes difficulties when trying to copy a line with a double click.
After inserting into the code, you have to delete parentheses.
In my opinion, this formatting option is more convenient and adapted for copy paste.


whatyouhide:
Thank you @dmitrydprog! 💟

PR #228, opened 2019-11-14
added context aware translation through pgettext, dpgettext, pngettext and dpngettext

Zurga:
As stated in the title, this pull request implements the context aware functionality. What follows are the biggest changes:

## Gettext.Compiler:
`lgettext/4` has been updated to `lgettext/5` to accept the `msgctxt` argument, the same goes for the `lngettext/7` function. I updated the tests to reflect these changes. 

`dpgettext` and `dpngettext` are now the final functions that are called by the other functions that have less functionality (dgettext, pgettext, gettext and ngettext, pngettext dngettext respectively). Defaults are applied as per the specification, a non-existent context is represented as `nil`. 

`dpgettext_noop` returns a tuple of {msgid, msgctxt}, `dpngettext_noop` now returns a tuple of {{msgid, msgid_plural}, msgctxt}. I tried a flat tuple for the `dpngettext_noop` function but was given a cryptic AST error... This is my pragmatic solution to that error.

## Gettext.Extractor:
`extract/5` has been changed to `extract/6` by the addition of the msgctxt argument. This is passed to the Translation struct properly. 
`merge_translations/2` has been modified to set the `msgctxt` of the new Translation struct.

I also updated the Context section of the documentation to reflect the changes. 
Please let me know what is missing/wrong so I can add/change it.


lib/gettext/extractor.ex
```diff
@@ -157,10 +159,11 @@ defmodule Gettext.Extractor do
   end
 
   defp create_translation_struct(msgid, msgctxt, file, line, extracted_comments) do
-    IO.puts msgctxt
+    IO.puts(msgctxt)
```
Wijnand: I think you left some debugging code :-)

Zurga: Thanks, I missed that one.

lib/gettext.ex
```diff
@@ -341,13 +341,10 @@ defmodule Gettext do
...
-  complexity of Gettext and would increase the complexity of the implementation
-  of Gettext for Elixir, and for this reason we decided to not support them. The
-  problem they try to solve can still be solved just using domains: for example,
-  one could have the `default-imperative_verbs` domain and the `default-nouns`
-  domain and use the `d(n)gettext` family of macros/functions, and the final
-  result would be similar
+  "nouns" context as to avoid ambiguity. The functions that handle contexts are 
+  are `pgettext`, `dpgettext`, `pngettext` and `dpngettext`. 
+  `dpgettext` can be read as d(omain)p(articular)gettext. The same reading rules
+  apply for `dnpgettext`.
```
josevalim:
```suggestion
  apply for `dpngettext`.
```

lib/gettext.ex
```diff
@@ -710,27 +741,69 @@ defmodule Gettext do
...
+        use Gettext, otp_app: :my_app
+      end
+
+      Gettext.dngettext(MyApp.Gettext, "errors", "Error", "%{count} errors", 3)
+      #=> "3 errori"
+      Gettext.dngettext(MyApp.Gettext, "errors", "Error", "%{count} errors", 1)
+      #=> "Errore"
+
+  """
+
```
josevalim:
```suggestion
```

lib/gettext.ex
```diff
@@ -634,6 +631,46 @@ defmodule Gettext do
...
+      end
+
+      Gettext.put_locale(MyApp.Gettext, "it")
+
+      Gettext.dpgettext(MyApp.Gettext, "errors", "user error", "Invalid")
+      #=> "Non valido"
+      
+      Gettext.dgettext(MyApp.Gettext, "errors", "signup form", "%{name} is not a valid name", name: "Meg")
+      #=> "Meg non è un nome valido"
+  """
```
josevalim: Please move the docs to the first occurrence of the function clause (i.e. line 634).

lib/gettext.ex
```diff
@@ -664,18 +701,12 @@ defmodule Gettext do
...
+  def dgettext(backend, domain, msgid, bindings \\ %{}) do
+    dpgettext(backend, domain, nil, msgid, bindings)
   end
 
-  def dgettext(backend, domain, msgid, bindings)
-      when is_atom(backend) and is_binary(domain) and is_binary(msgid) and is_map(bindings) do
-    locale = get_locale(backend)
-    result = backend.lgettext(locale, domain, msgid, bindings)
-    handle_backend_result(result, backend, locale, domain, msgid)
+  def pgettext(backend, msgctxt, msgid, bindings \\ %{}) do
```
josevalim: We need docs and specs for this function.

lib/gettext/compiler.ex
```diff
@@ -111,45 +126,100 @@ defmodule Gettext.Compiler do
...
             __MODULE__,
             domain,
+            msgctxt,
             {msgid, msgid_plural},
             Gettext.Compiler.get_and_flush_extracted_comments()
           )
         end
 
-        {msgid, msgid_plural}
+        {{msgid, msgid_plural}, msgctxt}
```
josevalim: Please make it return a three element tuple instead.

lib/gettext/compiler.ex
```diff
@@ -373,14 +477,17 @@ defmodule Gettext.Compiler do
        ) do
     msgid = IO.iodata_to_binary(t.msgid)
     msgstr = IO.iodata_to_binary(t.msgstr)
+    msgctxt = if !is_nil(t.msgctxt), do: IO.iodata_to_binary(t.msgctxt), else: nil
```
josevalim:
```suggestion
    msgctxt = t.msgctxt && IO.iodata_to_binary(t.msgctxt)
```

lib/gettext/compiler.ex
```diff
@@ -401,6 +508,7 @@ defmodule Gettext.Compiler do
     msgid = IO.iodata_to_binary(t.msgid)
     msgid_plural = IO.iodata_to_binary(t.msgid_plural)
     msgstr = Enum.map(t.msgstr, fn {form, str} -> {form, IO.iodata_to_binary(str)} end)
+    msgctxt = if !is_nil(t.msgctxt), do: IO.iodata_to_binary(t.msgctxt), else: nil
```
josevalim:
```suggestion
    msgctxt = t.msgctxt && IO.iodata_to_binary(t.msgctxt)
```

lib/gettext/compiler.ex
```diff
@@ -60,48 +60,63 @@ defmodule Gettext.Compiler do
...
           )
         end
 
-        msgid
+        {msgid, msgctxt}
+      end
+
+      defmacro dgettext_noop(domain, msgid) do
+        quote do
+          unquote(__MODULE__).dpgettext_noop(unquote(domain), nil, unquote(msgid))
```
josevalim: Note to myself (@josevalim) to refactor after merge (as it was not introduced as part of this PR): remove the double macro dispatch.

lib/gettext/compiler.ex
Zurga: I tried that first, but I got an error saying that this is not a correct AST node. My pragmatic approach was then to turn it into this tuple. See the error message below after running `mix test`:
```elixir
== Compilation error in file test/gettext_test.exs ==
** (CompileError) test/gettext_test.exs: invalid quoted expression: {"You have one message, %{name}", "You have %{count} messages, %{name}", nil}

Please make sure your quoted expressions are made of valid AST nodes. If you would like to introduce a value into the AST, such as a four-element tuple or a map, make sure to call Macro.escape/1 before
    expanding macro: GettextTest.Translator.dpngettext_noop/4
    test/gettext_test.exs:486: GettextTest."test dngettext/5"/1
    expanding macro: GettextTest.Translator.dpngettext/6
    test/gettext_test.exs:486: GettextTest."test dngettext/5"/1
    expanding macro: GettextTest.Translator.dngettext/5
    test/gettext_test.exs:486: GettextTest."test dngettext/5"/1
    (ex_unit) expanding macro: ExUnit.Assertions.assert/1
    test/gettext_test.exs:486: GettextTest."test dngettext/5"/1
    (elixir) lib/kernel/parallel_compiler.ex:237: anonymous fn/4 in Kernel.ParallelCompiler.spawn_workers/7
```

lib/gettext.ex
Zurga: Will do!

lib/gettext/compiler.ex
josevalim: Make it return this then:

```suggestion
        {:{}, [], [msgid, msgid_plural, msgctxt]}
```

Which is the AST for a  three element tuple. :)

josevalim: If the above doesn't work, then keep it as is and i will take a look later!

Zurga: Will do!

josevalim:
Btw, please let me know when this is ready for one final review. :)

Zurga:
@josevalim I will, thank you for the comments and the AST node solution. I learned something!

Zurga:
I think this pull request is ready for a final review.

lib/gettext/compiler.ex
Zurga: Would this allow for documentation of the Gettext macros directly? The ExDoc Autolinker now incorrectly links some documentation to the functions that reside in the Gettext module.
I see that Ecto uses Module.register_attribute/3 in their Schema module to add the macros, is the best way to add macros that require documentation?

josevalim: Not really, the comment above is mostly about refactoring things a bit. :)

Zurga: Ah okay, I thought you were also referencing the call to macros() in the @__before_compile__. Maybe the Gettext macros could go in a different module, then they can easily be documented and are separated from the internal API

Zurga:
@josevalim I think it is ready for review. Anything you find, I will fix later.

whatyouhide:
Thank you @Zurga! 💟

Zurga:
Your very welcome! @Wijnand at Tell Charlie sponsored the development.

Issue #205, opened 2019-07-01
Dialyzer errors on `Gettext.PO.parse_string`


hauleth:
Currently Dialyzer states that:

```elixir
{:ok, _} = Gettext.PO.parse_string(string)
```

will always fail, as this function always return `{:error, non_neg(), binary()}`.

josevalim:
Thanks for the report. We don't use dialyzer ourselves, so a PR is welcome!

hauleth:
@josevalim you mean PR that fix the issue or that will run Dialyzer on CI?

josevalim:
That fixes the issue, thanks!
-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

josevalim:
Ping. :D

hauleth:
Hell, I have forgotten about it. I will try to write something.

whatyouhide:
@hauleth we don't really use Dialyzer. If you get time to work on something we'd appreciate it. Closing the issue in the meantime, thank you!

kenny-evitt:
I just created a pull request that resolves this warning, and several others:

 - [Fix spec and update doc for one function by kenny-evitt · Pull Request #230 · elixir-gettext/gettext](https://github.com/elixir-gettext/gettext/pull/230)

@whatyouhide @josevalim Are you interested in using Dialyzer for this project?

josevalim:
We are not, as we usually don't have Dialyzer in our workflows. Thanks for asking though!

PR #230, opened 2019-11-15
Fix spec and update doc for one function

kenny-evitt:
This resolves all Dialyzer warnings for the `Gettext.PO` module.

Issue elixir-gettext/gettext#205
Issue kenny-evitt/gettext#1


whatyouhide:
Great catch, thanks @kenny-evitt! 💟

PR #231, opened 2019-12-17
Fix typo in default locale docs

char0n:
Thanks from the library. While reading docs I've noticed there is in-proper `with` inside the sentence.


whatyouhide:
Thank you @char0n! 💟

char0n:
No problem. My pleasure! ;]

Issue #233, opened 2020-01-11
Call to dpgettext breaks contract


alessandrousseglioviretta:
Dialyxir complains that the call to `dpgettext`in the `dgettext` function in `lib/gettext.ex`

```
def dgettext(backend, domain, msgid, bindings \\ %{}) do
    dpgettext(backend, domain, nil, msgid, bindings)
end
```

breaks the following contract:

`@spec dpgettext(module, binary, binary, binary, bindings) :: binary`

Being  `nil`not `binary`, it seems that the contract is broken indeed.

whatyouhide:
@alessandrousseglioviretta could you link to the code that breaks the contract? I'll be on mobile only for a couple days 😁, so it would help!

josevalim:
A pull request would be even better! Generally speaking we don’t use Dialyzer, so unless someone provides a fix, it is unlikely we will tackle it ourselves. :)

alessandrousseglioviretta:
I fixed the issue with the call to `dpgettext`, but there are still a few issues, like  `Extractor.create_translation_struct` that returns a _struct that differs from the success typing arguments_, which do not seem trivial to fix. I think it's worthwhile to fix them, but I'll need some support.

whatyouhide:
@alessandrousseglioviretta could you provide more insight on your fix and on the issues you mentioned?

alessandrousseglioviretta:
changed the spec for `dpgettext` from:

`@spec dpgettext(module, binary, binary, binary, bindings) :: binary`

to:

`@spec dpgettext(module, binary, binary | nil, binary, bindings) :: binary`

and , in type definition for `Gettext.PO.Translation` and `Gettext.PO.PluralTranslation`, turned

`po_source_line: pos_integer`

to

`po_source_line: pos_integer | nil`

josevalim:
Can you please send a PR with those fixes? Thank you!

alessandrousseglioviretta:
[...]

PR #235, opened 2020-01-11
dialyzer warnings are gone

alessandrousseglioviretta:
see https://github.com/elixir-gettext/gettext/issues/233#issuecomment-573359495


lib/gettext/po/plural_translation.ex
```diff
@@ -31,25 +31,25 @@ defmodule Gettext.PO.PluralTranslation do
 
   @type t :: %__MODULE__{
           msgid: [binary],
-          msgid_plural: [binary],
+          msgid_plural: [binary] | nil,
           msgstr: %{non_neg_integer => [binary]},
           msgctxt: [binary] | nil,
-          comments: [binary],
+          comments: [binary] | [],
```
josevalim: `[binary]` also includes empty list.

lib/gettext/po/plural_translation.ex
```diff
@@ -31,25 +31,25 @@ defmodule Gettext.PO.PluralTranslation do
...
-          po_source_line: pos_integer
+          po_source_line: pos_integer | nil
         }
 
   @enforce_keys [:msgid, :msgid_plural]
 
   defstruct msgid: nil,
             msgid_plural: nil,
-            msgstr: nil,
+            msgstr: [""],
```
josevalim: This and the ones below should be empty lists since `[binary]` also includes `[]`.

lib/gettext/po/plural_translation.ex
```diff
@@ -31,25 +31,25 @@ defmodule Gettext.PO.PluralTranslation do
...
             msgctxt: nil,
-            comments: [],
-            extracted_comments: [],
+            comments: [""],
+            extracted_comments: [""],
             references: [],
             flags: MapSet.new(),
-            po_source_line: nil
+            po_source_line: 1
```
josevalim: This can continue as `nil` since you added `nil` support above.

lib/gettext/po/translation.ex
```diff
@@ -30,23 +30,23 @@ defmodule Gettext.PO.Translation do
...
           flags: MapSet.t(),
-          po_source_line: pos_integer
+          po_source_line: pos_integer | nil
         }
 
   @enforce_keys [:msgid]
 
   defstruct msgid: nil,
-            msgstr: nil,
+            msgstr: [""],
```
josevalim: You need to either add `nil` to the typespec or make this an empty list. There is no reason to change both. :D FWIW, I would prefer to change the default here. Thanks!

lib/gettext/po/plural_translation.ex
alessandrousseglioviretta: unfortunately Dialyzer needs a `pos_integer`, `nil` won't work.
To confirm that, simply add the line `translation = Map.put(translation, :po_source_line, 1)` in `Extractor.extract`, which will make Dialyzer happy. I guess that the case of `po_source_line` being `nil`is not handled by the code, but I could not find any evidence.

lib/gettext/po/plural_translation.ex
alessandrousseglioviretta: that's right

lib/gettext/po/plural_translation.ex
alessandrousseglioviretta: that's right

lib/gettext/po/translation.ex
alessandrousseglioviretta: I've set the default to an empty list: `msgstr: []`

lib/gettext/po/plural_translation.ex
josevalim: So let's keep this as `1` and remove the `nil` from the typespec. Also please make sure to rewrite all `[""]` as `[]`. Thanks!

alessandrousseglioviretta: done

josevalim:

PR #239, opened 2020-01-18
Support configuration of locales by environment

### LINKED ISSUE #238: Compilation time with many locales
In my application, gettext compile takes around 10 minutes in developer's machines (docker inside mac, so i don't expect it to fly. On CI and production it is several minutes, but not as much as 10)
I have around 20 locales with around 6000 strings each, and I will keep adding more, so I need a way to decrease developer's time spent on recompilations.
I'm not as concerned about paying this price on production or CI, but for tests and development, i'm trying to find a way to use less locales.

My first thought was to have a `priv/gettext_dev` folder for development, and the normal `priv/gettext` for production. But that would force me to an extra step each time we run gettext_extract, as it's ran in the development environment, copying the .pot files from one folder to the other.
Second thought is to have a way to restrict the number of locales processed in development environment. Would you accept a PR with a whitelist of locales to compile, so it can be configured per environment?

Do you have any other idea or out of the box solution that i'm missing?

epilgrim:
Added a configuration option `allowed_locales` to restrict the locales that gettext bundles in the backend.

This patch introduces slightly more changes than I was expecting. 
In the previous version, the filesystem was read twice. One to discover the locales, and another to load the po files.
If I had kept this approach, I should have done the filtering of locales twice, with different inputs.
In the current approach, I reach for the filesystem once, and filter the PO files there based on the configuration, returning the path, locale and domain in the same data structure.
This structure is then cascaded to the different compilers

closes #238

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.18.0
+
+* Add `:allowed_locales` to restrict the locales bundled in the backend
+
 ## v0.17.4
 
 * Do not change the return types of `*_noop` macros (regression in v0.17.2 and v0.17.3)
--- a/lib/gettext/compiler.ex
+++ b/lib/gettext/compiler.ex
@@ -33,7 +33,8 @@ defmodule Gettext.Compiler do
     priv = Keyword.get(opts, :priv, @default_priv)
     translations_dir = Application.app_dir(otp_app, priv)
     external_file = String.replace(Path.join(".compile", priv), "/", "_")
-    known_locales = known_locales(translations_dir)
+    known_po_files = known_po_files(translations_dir, opts)
+    known_locales = Enum.map(known_po_files, & &1[:locale]) |> Enum.uniq()
 
     default_locale =
       opts[:default_locale] || quote(do: Application.fetch_env!(:gettext, :default_locale))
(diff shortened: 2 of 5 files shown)


lib/gettext/compiler.ex
```diff
@@ -629,12 +629,19 @@ defmodule Gettext.Compiler do
...
   end
 
-  # Returns all the locales in `translations_dir` (which are the locales known
-  # by the compiled backend).
-  defp known_locales(translations_dir) do
+  # Returns the known the PO files in `translations_dir` with their locale and domain
+  # If allowed_locales is configured, it removes all the PO files that do not belong
+  # to those locales
+  defp known_po_files(translations_dir, opts) do
     case File.ls(translations_dir) do
```
epilgrim: I kept the `File.ls` here to maintain the same interface as before when there is the need to raise the exception

test/gettext_test.exs
```diff
@@ -700,7 +715,8 @@ defmodule GettextTest do
 
   test "known_locales/1: returns all the locales for which a backend has PO files" do
     assert Gettext.known_locales(Translator) == ["it"]
-    assert Gettext.known_locales(TranslatorWithCustomPriv) == ["it"]
+    assert Gettext.known_locales(TranslatorWithCustomPriv) == ["es", "it"]
```
epilgrim: I used the same custom `priv` in the new translator. Let me know if I should use a different priv, or test the same priv but instead of adding a new locale, configure the translator not to allow any locale (`allowed_locales: []`)

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.18.0
+
+* Add `:allowed_locales` to restrict the locales bundled in the backend
```
epilgrim: if you can think of a better name, let me know. I was also toying with `whitelisted_locales`, `bundled_locales`

PR #240, opened 2020-01-21
Ensure that touching the PO file happens after the manifest

epilgrim:
There is a chance that the previous assertion leaves the timestamps in
the following state:

  | File                              | Timestamp
  ------------------------------------|------------------------------
  | Manifest .compile_tmp_gettext_bar | {{2020, 1, 21}, {17, 49, 57}}
  | Manifest .compile_tmp_gettext_foo | {{2020, 1, 21}, {17, 49, 57}}
  | PO foo/en/LC_MESSAGES/hello.po    | {{2020, 1, 21}, {17, 49, 56}}

which means that when touching `hello.po` to have a newer timestamp, ends
up with the same timestamp as the manifest. This marks the whole manifest as "not stale".

With this patch, i'm forcing the touch of the file to be later than the
oldest of the timestamps of the PO or the manifest.

This should fix the error https://travis-ci.org/elixir-gettext/gettext/jobs/638875306?utm_medium=notification&utm_source=github_status

image


josevalim:

PR #244, opened 2020-02-17
Allow sorting strings by msgid

angelikatyborska:
I have been working on one Phoenix app with a few people for 1 year now. Every time we happened to add strings to the same Gettext domain on two different feature branches, we would get merge conflicts because the standard behavior of Gettext is to append strings to the end of the file. This is a minor inconvenience for a single file, but a bit annoying if you have more than 5 languages.

We came to the conclusion that we might decrease how often conflicts happen if we changed the default behavior - instead of appending strings at the end, which guarantees conflicts every time, we want to sort them alphabetically by `msgid`.

We used that approach for a few months now, and it does work for decreasing the number of merge conflicts. Thus I want to suggest it as a option for everyone :).

The downside is that the files might change the first time you run `mix gettext.extract` after you resolved conflicts manually and didn't pay attention to ordering the strings.

### DIFF
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -514,7 +514,7 @@ defmodule Gettext do
     * `:sort_by_msgid` - a boolean that modifies the sorting behavior.
       By default, the order of existing translations in a POT file is kept, and new
       translations are appended to the file. If `:sort_by_msgid` is set to true,
-      existing and new translations will be mixed and be sorted alphabetically by msgid.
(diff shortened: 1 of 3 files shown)


angelikatyborska:
Oops, I am a bit surprised by those test failures. It looks to me like the tests are already broken on master and it's not related to my changes.

josevalim:

josevalim:
I have fixed the test failures now, thank you!

PR #254, opened 2020-06-07
Configurable Default Domain

### LINKED ISSUE #253: Configurable default domain?
I’ve been working with a couple of different JS implementations of gettext (before rolling a fork of one of them to work better with Vue), and it appears that _most_ JS implementations use `messages` as their default domain—and have an easy way to change the active default domain.

As best as I can tell, this would be set by calling `textdomain(domainname)` in the C version of gettext. What I think would be useful would be providing a `domain` value to `use Gettext`:

```elixir
defmodule MyApp.Gettext do
  use Gettext, otp_app: :my_app, domain: "messages"
end
```

I can work up a PR for this if that makes sense.

halostatue:
Resolves #253

A given backend can be configured to have a default domain with the option
`:default_domain`, as in:

```elixir
config :gettext, MyApp.Gettext, default_domain: "messages"

defmodule MyApp.Gettext do
  use Gettext, otp_app: :my_app, default_domain: "messages"
end
```

When applied, this means that instead of `MyApp.Gettext.gettext("foo")` being
equivalent to `MyApp.Gettext.dgettext("default", "foo")`, it is now equivalent
to `MyApp.Gettext.gettext("messages", "foo")`.

### DIFF
--- a/lib/gettext/compiler.ex
+++ b/lib/gettext/compiler.ex
@@ -11,6 +11,7 @@ defmodule Gettext.Compiler do
   require Logger
 
   @default_priv "priv/gettext"
+  @default_domain "default"
   @po_wildcard "*/LC_MESSAGES/*.po"
 
   @doc false
@@ -38,6 +39,9 @@ defmodule Gettext.Compiler do
 
     default_locale =
       opts[:default_locale] || quote(do: Application.fetch_env!(:gettext, :default_locale))
+    default_domain = opts[:default_domain] || @default_domain
+
+    Module.put_attribute(env.module, :gettext_default_domain, default_domain)
 
     quote do
       @behaviour Gettext.Backend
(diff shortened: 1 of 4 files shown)


halostatue:
There are parts of this that I’m not particularly fond of, but don’t see an easy way around. Because of `Gettext.Compiler.expand_to_binary/4`, it’s necessary to have `domain = __gettext__(:default_domain)` in each compiled macro, which results in what feels like an opportunity for maintenance drift.

As mentioned in #253, I’m not sure of two things:

1. Global or per-backend? That is, should there be `config :gettext, default_domain: "messages"`, or should it just exist in the backends? I’ve assumed the latter.
2. Just in the macros, or in the global calls, too? That is, should `Gettext.gettext(MyApp.Gettext, "foo")` be the same as `Gettext.dgettext(MyApp.Gettext, "default", "foo")` or `Gettext.dgettext(MyApp.Gettext, MyApp.Gettext.__gettext__(:default_domain), "foo")`. I’ve assumed the former.

josevalim:

lib/gettext/compiler.ex
```diff
@@ -39,6 +40,10 @@ defmodule Gettext.Compiler do
     default_locale =
       opts[:default_locale] || quote(do: Application.fetch_env!(:gettext, :default_locale))
 
+    default_domain = opts[:default_domain] || @default_domain
+
+    Module.put_attribute(env.module, :gettext_default_domain, default_domain)
```
halostatue, line 45: Well, I didn’t meant to leave _that_ in. It doesn’t seem to be causing any issues, though.

josevalim: Good point. removed on master :)

PR #262, opened 2020-07-31
Improve parallelism when compiling modules

epilgrim:
Instead of loading everything into memory at once, I'm compiling the locales one by one in a stream, and then passing it to the parallel compiler.

In a project of mine, compile time goes from 2m 11s to 1m25s.
Memory consumption also goes down, from 3.35G to 2.56G

I'm aware the resulting code is not the most elegant. If you have ideas of better takes to do the same, I'm happy to try them. This is the minimal version that accomplishes the task of reducing memory and run time. Other takes would require more involved changes


josevalim:

Issue #266, opened 2020-10-23
Bindings interpolation with string keys


kpanic:
Hello everyone, I was wondering if it would be possible to have bindings in this format:

```elixir
interpolatable = Interpolation.to_interpolatable("%{a} %{b} %{c}")

assert Interpolation.interpolate(interpolatable, %{"a" => 1, "b" => :two, "c" => "thr ee"},) ==
               {:ok, "1 two thr ee"}

```

If not, is there a technical reason?

Thanks for the library!

josevalim:
It should be possible but can you tell us a bit more about your use case?

kpanic:
In my use case we have dynamic translations (that can be updated in time)
AFAIK in the current `interpolate/2` function only supports map with atom keys and I have to convert strings to atoms risking that the atom table will grow and crash the BEAM

josevalim:
@kpanic the string is dynamic but the map you pass to it is known upfront, right?

kpanic:
No, in our use case we can't. The map can be dynamic (from user input/spefication)

josevalim:
Closing in favor of PR discussion.

PR #267, opened 2020-10-23
Binding interpolation with string keys

kpanic:
Hi guys,

This is a super **work in progress** to get feedback from you folks.
It tries to address (partially for the moment) issue https://github.com/elixir-gettext/gettext/issues/266

For the bindings:

* I am introducing a `:mode` that it is `:atom` by default
* In `to_interpolatable` function I am marking (only in the case of non `:atom` maps) the binding with `%{}` since I did not found a way to determine the string binding

Currently this implementation is kind of dirty I know, I am open to feedback if this approach is worth to carry on.
I was thinking to add the `:mode` as a macro parameter in `compiler.ex` to be able to specify it (not in this PR yet)

Thanks a lot!

### DIFF
--- a/lib/gettext/interpolation.ex
+++ b/lib/gettext/interpolation.ex
@@ -20,17 +20,19 @@ defmodule Gettext.Interpolation do
       ["Empties %{} stay empty"]
 
   """
+  def to_interpolatable(string, opts \\ [mode: :atom])
+
   @spec to_interpolatable(String.t()) :: interpolatable
-  def to_interpolatable(string) do
+  def to_interpolatable(string, opts) do
     start_pattern = :binary.compile_pattern("%{")
     end_pattern = :binary.compile_pattern("}")
 
     string
-    |> to_interpolatable(_current = "", _acc = [], start_pattern, end_pattern)
+    |> to_interpolatable(_current = "", _acc = [], start_pattern, end_pattern, opts)
     |> Enum.reverse()
   end
 
-  defp to_interpolatable(string, current, acc, start_pattern, end_pattern) do
+  defp to_interpolatable(string, current, acc, start_pattern, end_pattern, opts) do
     case :binary.split(string, start_pattern) do
       # If we have one element, no %{ was found so this is the final part of the
       # string.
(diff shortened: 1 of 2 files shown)


josevalim:
Hi @kpanic! This looks good to me but I would love to see how it will be used in the long run. That's because Interpolatable is a private APIs, so you shouldn't call it directly. But I assume the plan is to expose this at the higher level at some point? Thanks!

kpanic:
hi @josevalim thanks for the feedback. I'll continue this PR in this direction then.
Yes, the plan would be to expose it at high level.
Thank you!

josevalim:
Actually, I think there is an easier approach. Hold on a bit. :)

josevalim:
No, nevermind. Please go ahead if you have any ideas. :)

josevalim:
Btw, are your dynamic translations never compiled? Or are they written to .po files and then eventually compiled?

kpanic:
the translations are kept in memory @josevalim 
we use `Gettext.PO.parse_string(pot_payload)` to parse a pot file to obtain the dynamic translations

josevalim:
I see. I actually think we should not go ahead with this then. Gettext.Interpolatable is private API and given that they are in memory, you are likely not using the higher-level `gettext` functions either. Your lookups/compilation for memory is very distinct than the one in gettext that I think you should probably roll your own interpolatable (it is relatively small) that emits a data structure that is semantically better for you.

My suggestion is probably to implement a library called dyngettext, that depends on this one for the PO parsing and stuff, but rolls your own interpolatable and other helpers.

kpanic:
Thanks @josevalim for elaborating on this! I will think about it

kpanic:
Hey @josevalim morning :)
I was thinking that I could override the `handle_*` callbacks in the `Gettext` macro to swap transparently them under the hood.
In particular, instead of importing https://github.com/elixir-gettext/gettext/blob/master/lib/gettext.ex#L606
I would:         
`import Gettext.DynamicInterpolation, only: [to_interpolatable: 1, interpolate: 2]`
in a potential `dyngettext` library.
How this approach sounds like?

josevalim:
Would you ever have static translations? The ones from .po files?

kpanic:
nope, they change in time (not that fast) but they change.
We are using an external service for translations and we can fetch the `pot` file from this service

josevalim:
Which others benefits are you getting from gettext them? Do you still use it to build the POTs?

kpanic:
No, we don't build the POTs directly @josevalim 
We are using Gettext to parse the POTs and to handle translations with the `Gettext.gettext/3` functions

kpanic:
Regarding the `po`(s) they come from this external API, for the record

josevalim:
Yeah, so my suggestion is to roll your own and not depend on Gettext, except for PO parsing. I think trying to make things fit Gettext will just make things slower and less than ideal. :) (and there is very little to reuse outside of the parsing anyway).

josevalim:
Closing per the above, thanks! :)

PR #273, opened 2021-02-16
Fix typo in conflicting backends warning

roodion:


josevalim:

roodion:
@josevalim may i ask your oppinion on this weired behaviour: 

While this error-message is helpful, in my specific case it's still not much i understand.
At my project i got

```bash
warning: the Gettext backend MyUmbrellaApp.Web has the same :priv directory as MyUmbrelaApp.Web, which means they will override each other.
```

I believe it's because i have following module

```
defmodule MyUmbrellaApp.Translator do
  use Gettext, otp_app: :web
end
```
which i import in multiple places within an app as i need that.  
Am i using lib in wrong way and does it make sense to add another warning with good advice about correct Gettext usage? 
Also, as a result i have duplicated reference_comments in default.pot file 

Thanks for reading! 
Any help with that would be highly appreciated.

josevalim:
You probably have two different Gettext backends in your app pointing to the same place, you probably only need one. Search for `use Gettext` in your project.

roodion:
@josevalim you are absolutely right, i have multiple backends in my app which i structured [according to the doc](https://hexdocs.pm/gettext/Gettext.html#with_locale/3)

```elixir
defmodule MyApp.Backend1 do
  use Gettext, otp_app: :backend_1
end

#: and using it in different modules, like: 
defmodule MyApp.Page
  import MyApp.Backend1
  def do_smth do
    Gettext.with_locale(MyApp.Backend1, language, fn ->
      message = Gettext.gettext(MyApp.Backend1, msgid)
    end)
  end
end

defmodule MyApp.Report
  import MyApp.Backend1
  def header do
    [gettext_noop("Name"), gettext_noop("City"), gettext_noop("State"), gettext_noop("Country")]
  end
end
```

And same structure for Backend2 in another part of app
Is this incorrect usage of GetText or i have possible issue with imports modules in my app?

josevalim:
The issue is not with imports. The issue is having multiple backends pointing to the same priv. It is still not clear why you need it.

PR #276, opened 2021-06-07
Change references from "country" to "territory"

kipcole9:
"country" refers to a political entity which can cause conflict when territories are disputed.

A list of disputed territories is available at https://en.wikipedia.org/wiki/List_of_territorial_disputes

This commit uses the term "territory" rather than "country" to reflect that locales serve a purpose
to reflect user preferences, not to express a political agreements.

* Annex A to ISO/IEC 15897:1998(E) also refers to "territory" for submissions.

* ISO3166 which lists "country codes" says for "TW" (Taiwan) that ISO recognises Taiwan as "Taiwan, Province of China". This view is not widely accepted by the people who live in Taiwan showing how challenging it is to reflect
both cultural preferences and manage political realities.

Using the term "territory" tends to nullify these issues and focus on the reason that a locale definition exists.

CLDR (https://cldr.unicode.org) uses the terms "territory" or "region" interchangeably for this reason.

### DIFF
--- a/lib/gettext/backend.ex
+++ b/lib/gettext/backend.ex
@@ -12,16 +12,16 @@ defmodule Gettext.Backend do
 
   For example, if something like this is called:
 
-      MyApp.Gettext.gettext("Hello %{name}, welcome to %{territory}", name: "Jane", territory: "Italy")
+      MyApp.Gettext.gettext("Hello %{name}, your favorite color is %{color}", name: "Jane", color: "blue")
 
   and our `it/LC_MESSAGES/default.po` looks like this:
 
-      msgid "Hello %{name}, welcome to %{territory}"
-      msgstr "Ciao %{name}, benvenuto in %{cowntry}" # (typo)
+      msgid "Hello %{name}, your favorite color is %{color}"
+      msgstr "Ciao %{name}, il tuo colore preferito è %{colour}" # (typo)
 
   then Gettext will call:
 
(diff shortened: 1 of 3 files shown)


whatyouhide:
I like this, good catch. My only question: should we mention what you say in the PR description in the docs too? I’m afraid many folks won't have the context to understand what a "territory" is when they're used to it being referred to as "country". Thoughts?

lib/gettext/backend.ex
```diff
@@ -12,11 +12,11 @@ defmodule Gettext.Backend do
 
   For example, if something like this is called:
 
-      MyApp.Gettext.gettext("Hello %{name}, welcome to %{country}", name: "Jane", country: "Italy")
+      MyApp.Gettext.gettext("Hello %{name}, welcome to %{territory}", name: "Jane", territory: "Italy")
```
josevalim, line 15:
```suggestion
      MyApp.Gettext.gettext("Hello %{name}, your favorite color is %{color}", name: "Jane", color: "blue")
```

josevalim:
@whatyouhide I changed the example slightly. WDYT?

whatyouhide:
@josevalim yeah totally agree, easier this way. I still think we mention "territory" a few times in the docs and I’m afraid the question I asked above still holds 🤔

josevalim:
We only mention it now on locale_TERRITORY and it is immediately followed by an example, en_US, so I think we are good.

whatyouhide:
Thanks @kipcole9! 💟

PR #278, opened 2021-08-10
Use UDN backup for Localization and Plurals.

czwartyeon:
Hi!

Hope you're well.

It appears that `https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_and_Plurals` is no longer a valid link. This PR suggests using a UDN backup of the `Localization and Plurals` article.

👋


josevalim:

PR #281, opened 2021-08-20
Make dngettext/6 bindings optional

ibarchenkov:
According to [the examples](https://hexdocs.pm/gettext/Gettext.html#dngettext/6-examples) `dngettext/6`'s last argument `bindings` is supposed to be optional.
Please let me know If I need to leave the function as is and edit the examples instead.


josevalim:

Issue #286, opened 2021-09-11
New mix task to check for missing translations


akoutmos:
Opening up this issue to gauge interest from the maintainers in a new mix task for gettext that will check to see if there are any missing translations. Something like `mix format --check-formatted` but for translations. This can be useful in CI to ensure that the PO files are up to date prior to deployment. I am currently doing this with adhoc shell scripts...but I think this would be a useful tool to have directly in gettext.

If there is interest in this, I can put together an initial implementation :).

As always, thanks for all the effort everyone has put into gettext!! Really appreciate it!!

josevalim:
I believe one of the reasons we haven't implemented this is because existing tools used by translators to translate gettext already include this functionality. Have you tried this route? Or am I misremembering?

akoutmos:
You can do something like follows to get information on whether a PO file is missing any translations:

```
$ msgfmt --statistics priv/gettext/en/LC_MESSAGES/default.po 

0 translated messages, 269 untranslated messages.
```

The check that I envision would be more about the creation of the POT file and would check to see if after adding new `gettext "Some new text"` statements in the code base, that an up to date POT file has been generated. Perhaps this would be a flag on `mix gettext.extract`? Something like `mix gettext.extract --check-template-file`?

josevalim:
Oh, you want a check to see if someone forgot to call `mix gettext.extract`? What about `--check-unextracted` or `--check-missing` or `--check-pending`?

akoutmos:
Oooo. I like `--check-unextracted`. I'll put something together in the next few days and open up a PR :).

whatyouhide:
That's awesome and great idea @akoutmos! Let us know if we can help with guidance on anything 🙃

akoutmos:
Will do! Thanks :)

akoutmos:
Closing this issue given that the PR has been merged. Thanks for the assistance everyone and the speedy reviews!

PR #287, opened 2021-09-16
Added --check-unextracted to extract mix task

akoutmos:
This PR addresses #286.

While still a work in progress (mostly test clean up that I want to do), I figured I would throw up the PR to get eyes on it sooner rather than later in case the approach I took was off.

Thanks in advance for reviewing the PR :)

### DIFF
--- a/lib/mix/tasks/gettext.extract.ex
+++ b/lib/mix/tasks/gettext.extract.ex
@@ -14,11 +14,12 @@ defmodule Mix.Tasks.Gettext.Extract do
   and `:priv` options given by Gettext modules when they call `use Gettext`. One
   POT file is generated for each translation domain.
 
-  If you would like to verify that your POT files are up to date and that there
-  are no missing gettext extractions, you can provide the `--check-unextracted`
-  flag:
+  If you would like to verify that your POT files are up to date with the
+  current state of the codebase, you can provide the `--check-up-to-date`
+  flag. Note that this validation will fail even when the same calls to gettext
+  only change location in the codebase:
 
-      mix gettext.extract --check-unextracted
+      mix gettext.extract --check-up-to-date
 
   It is possible to give the `--merge` option to perform merging
   for every Gettext backend updated during merge:
@@ -33,7 +34,7 @@ defmodule Mix.Tasks.Gettext.Extract do
 
   """
 
-  @switches [merge: :boolean, check_unextracted: :boolean]
+  @switches [merge: :boolean, check_up_to_date: :boolean]
 
   def run(args) do
     Application.ensure_all_started(:gettext)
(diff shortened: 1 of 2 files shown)


juanperi:
hey, this would be an awesome addition!
I have a question though. From the look at the PR, you are exactly matching the content of the previous file vs the new file, and failing if they don't.
Now, that is indeed correct, and very helpful. But what if there are no new translations, and the only thing that changed is the line in which a translation was taken from.
eg:
```
#, elixir-format
#: lib/app_web/templates/compare/index/_alternative_card.html.eex:49
msgid "# 1 Alternative"
msgstr ""
```
If we add a new line somewhere in that template, now the line will be 50, and your check will fail.

So, when should this test fail?:
1. If a new `gettext("content")` was added and it was not extracted
2. if any possible change (like line change) has happened and would make the POT file different

josevalim:
@juanperi I think the goal is still raise, because it means someone forgot to run the command. My understanding is that the goal of the flag is basically "did someone remember to run mix getttext.extract before committing?". :) But I am interested in @akoutmos' take too.

akoutmos:
I thought about this as well yesterday...and the conclusion that I came to was that while there are no new translations and the existing translations either changed lines/files, the POT file is still inaccurate with regards to the current state of the codebase. Hence why I made the PR the way that I did.

josevalim:
Hrm, maybe we should change the name? After all it is checking more than unextracted. Maybe `--check-up-to-date` or `--check-updated`?

akoutmos:
I guess `--check-unextracted` might be a bit misleading now that I think about it. I'd be ok with `--check-up-to-date` or perhaps `--check-current` for something shorter?

josevalim:
Any is fine. I am glad to leave it up to you. :)

juanperi:
hey, awesome we reached a consensus on the goal of the check. Thank you both.

I don't think that i would be able to use it for my particular case, where most of the time changes in the templates do not imply they are adding new translations, and although the line numbers are useful, they are mostly needed when new translations are made available. But I understand that strictness is better for a general check :)

Maybe the extra "configuration" flag for this check might be implemented later if you think it would be useful (i might work on that if there is interest)

akoutmos:
Finished the last bit of test clean up so all set on my end :).

josevalim (approved):

lib/mix/tasks/gettext.extract.ex
```diff
@@ -14,6 +14,13 @@ defmodule Mix.Tasks.Gettext.Extract do
   and `:priv` options given by Gettext modules when they call `use Gettext`. One
   POT file is generated for each translation domain.
 
+  If you would like to verify that your POT files are up to date with the
+  current state of the codebase, you can provide the `--check-up-to-date`
+  flag. Note that this validation will fail even when the same calls to gettext
+  only change location in the codebase:
+
+      mix gettext.extract --check-up-to-date
```
whatyouhide, line 22: Love the flag name. Awesome 💟

PR #288, opened 2021-09-28
Put each reference on its own line

### LINKED ISSUE #255: Apparent mishandling of reference comments on merge
What I’m reporting here is not unique to Elixir’s gettext implementation, as I have seen it with some version of `msguniq` from the official GNU tooling, and it may not be a bug, but it feels like one and it results in an unnecessarily large delta.

We ran `mix gettext.merge priv/gettext` and got a diff that looks like this:

```
diff --git a/priv/gettext/en_CA/LC_MESSAGES/translations.po b/priv/gettext/en_CA/LC_MESSAGES/translations.po
index 7b87ad0d..e61885ff 100644
--- a/priv/gettext/en_CA/LC_MESSAGES/translations.po
+++ b/priv/gettext/en_CA/LC_MESSAGES/translations.po
@@ -95,8 +95,7 @@ msgid "Account will expire on %{when}"
 msgstr "Account will expire on %{when}"
 
 #: assets/src/components/common/elements/modal/LinkChargeBuyerModal.vue:26
-#: assets/src/components/common/elements/tables/PurchaseHistoryDetails.vue:24
-#: assets/src/components/common/elements/tables/PurchaseTable.vue:41
+#: assets/src/components/common/elements/tables/PurchaseHistoryDetails.vue:24 assets/src/components/common/elements/tables/PurchaseTable.vue:41
 msgid "Account:"
 msgstr "Account:"
 
```

There are _dozens_ of cases like this. The problem is that it has taken two lines:

```gettext
#: assets/src/components/common/elements/tables/PurchaseHistoryDetails.vue:24
#: assets/src/components/common/elements/tables/PurchaseTable.vue:41
```

And turned them into one long line (much longer than 80 characters):

```gettext
#: assets/src/components/common/elements/tables/PurchaseHistoryDetails.vue:24 assets/src/components/common/elements/tables/PurchaseTable.vue:41
```

This is common enough with what I’ve been dealing with lately (both from the Elixir tools and `msguniq`) that I have a ex-command in my vim history: `:%s/\([^:]\) assets/\1^M#: assets/g` to fix it.

Gladear:
Fixes #255 

Update tests to match this new behavior

### DIFF
--- a/lib/gettext/po.ex
+++ b/lib/gettext/po.ex
@@ -27,8 +27,6 @@ defmodule Gettext.PO do
           file: nil | Path.t()
         }
 
-  @wrapping_column 80
-  @reference_comment_length String.length("#:")
   @bom <<0xEF, 0xBB, 0xBF>>
 
   @doc """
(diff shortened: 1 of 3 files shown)


josevalim:

Issue #290, opened 2021-10-26
Plugable Bindings Replacement


maennchen:
I would like to propose, to make the binding replacement in strings (`gettext("Hello %{name}", name: "World")`) configurable.

## Changes

1. When doing `use Gettext`, a replacement module can be specified
2. Gettext is providing a module that replaces the `%{...}` bindings
3. Another module can be supplied, that will provide a macro that receives the message as well as the bindings

## Reasoning

A lot of tooling requires specific formats of bindings. One example would be Localise:
https://docs.lokalise.com/en/articles/1400511-lokalise-placeholders

If I were able to for example provide a module, that implements ICU bindings and provide that directly to Gettext, that would be awesome.

## PR

I would like to propose a PR if there is interest in this feature.

maennchen:
@josevalim Awesome, thanks for your help!

PR #291, opened 2021-10-26
Pluggable Interpolation

maennchen:
Implementation of #290

Depends on #292  merge that PR first.


lib/gettext.ex
```diff
@@ -606,22 +610,22 @@ defmodule Gettext do
       defoverridable handle_missing_bindings: 2
 
       def handle_missing_translation(_locale, domain, msgid, bindings) do
-        import Gettext.Interpolation, only: [to_interpolatable: 1, interpolate: 2]
-
         Gettext.Compiler.warn_if_domain_contains_slashes(domain)
 
-        with {:ok, interpolated} <- interpolate(to_interpolatable(msgid), bindings),
+        with {result, _code_bindings} =
+               Code.eval_quoted(@interpolation.(msgid, Macro.escape(bindings))),
```
josevalim, line 616: `eval` is almost always an anti-pattern. I think a lot of the complexity introduced in this PR is because you have defined your contract to be a closure (an anonymous function). Instead, make the contract a behaviour, with two functions:

    compile_interpolate(:translation | :plural_translation, msg, bindings_ast)
    runtime_interpolate(msg, bindings_ast)

In here, you are going to call `@interpolation.runtime_interpolate`.

lib/gettext/compiler.ex
```diff
@@ -528,23 +548,28 @@ defmodule Gettext.Compiler do
 
       # If the msgid is the same as msgstr, simply return it without allocating
       # a new string.
-      ^msgid when keys == [] ->
```
josevalim, line 531: This optimization needs to be moved inside the interpolation module.

lib/gettext/compiler.ex
```diff
@@ -528,23 +548,28 @@ defmodule Gettext.Compiler do
...
-      _ ->
-        quote do
-          Kernel.unquote(kind)(
-            unquote(singular_fun)(unquote(msgctxt), unquote(msgid), var!(bindings))
-          ) do
-            unquote(compile_interpolation(msgstr, :translation))
-          end
-        end
+      _other ->
+        # TODO: How to do this nicely?
```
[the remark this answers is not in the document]
josevalim: You should be able to do it with `quote`, as it was done before.

lib/gettext/compiler.ex
```diff
@@ -622,67 +660,6 @@ defmodule Gettext.Compiler do
...
-
-    all_bindings_clause = quote do: (unquote(match) -> {:ok, unquote(interpolation)})
-
-    dynamic_interpolation_clause =
-      quote do
-        %{} -> Gettext.Interpolation.interpolate(unquote(interpolatable), var!(bindings))
-      end
-
-    clauses =
-      if keys == [:count] and translation_type == :plural_translation do
```
josevalim, line 653: You have discarded this optimization.

josevalim:
@maennchen I have a suggestion on how to best approach this problem. Instead of trying to make it generic and refactor at the same time, do this. Introduce `Gettext.Interpolation.compile_interpolate/3` and `Gettext.Interpolation.runtime_interpolate/2`, as mentioned in another comment, but without making it configurable. Move all of the code that you need inside those two functions and refactor the code as necessary.

Then, in another PR, you introduce the customization. This should reduce the amount of moving parts and therefore you should be able to leave most macros intact. WDYT? :)

maennchen:
@josevalim that's a good suggestion. Expect a pr soonish 🙂

lib/gettext/compiler.ex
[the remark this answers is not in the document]
maennchen: @josevalim With the separate module, this no longer applies.

lib/gettext.ex
maennchen: I've extracted this into #292. Further discussion in https://github.com/elixir-gettext/gettext/pull/292#discussion_r739878846.

lib/gettext/compiler.ex
maennchen: Re-added as https://github.com/elixir-gettext/gettext/pull/292/files#diff-3f91eab05a554c815d432e2899ed13dd20708be3f2019cc8c75077391ef20ffaR188

lib/gettext/compiler.ex
maennchen: Details also via https://github.com/elixir-gettext/gettext/pull/292/files#r739878846

lib/gettext.ex
```diff
@@ -496,6 +496,9 @@ defmodule Gettext do
       This option can be useful in development to reduce compile-time
       by compiling only a subset of all available locales.
 
+    * `:interpolation` - the name of a module that implements the
+      `Gettext.Interpolation` behaviour. Default: `Gettext.Interpolation.Elixir`
```
josevalim, line 500:
```suggestion
      `Gettext.Interpolation` behaviour. Default: `Gettext.Interpolation.Default`
```

I think Elixir is not very clear to mean in meaning.

lib/gettext/interpolation/elixir.ex
```diff
@@ -0,0 +1,220 @@
+# TODO: Better name?
+defmodule Gettext.Interpolation.Elixir do
```
maennchen, line 2: I really don't like the name. I was also thinking about `Default`, but I also don't think that name is better.

I thought calling it `Elixir` could be good, since the specification mentions various formats and this is basically the standard for elixir.

https://www.gnu.org/software/gettext/manual/gettext.html#The-Translator_0027s-View-1

lib/gettext.ex
maennchen: I was thinking about the same and wrote in https://github.com/elixir-gettext/gettext/pull/291#discussion_r740107319.

Do you still think `Default` is best after my other comment?

josevalim (commented):
Looks great! Just one comment regarding the name and we should be good!

lib/gettext.ex
josevalim: Hrm, the issue is that `%{...}` is not used anywhere in Elixir either, because Elixir doesn't have a format string function per se.

maennchen: @josevalim That's right as well... I'm not really happy with either of them, but I don't have any better ideas.

I'll change the name to `Default`.

lib/gettext/interpolation/elixir.ex
maennchen: Fixed in 58df32156fe4eeb0e7a1140616dfb5ada6f7a30e

lib/gettext.ex
maennchen: Fixed in 58df32156fe4eeb0e7a1140616dfb5ada6f7a30e

josevalim:

PR #292, opened 2021-10-31
Refactor Interpolation to be more self contained

maennchen:
This is a preparation to build pluggable interpolation backends in the future.

Follow up from https://github.com/elixir-gettext/gettext/pull/291#issuecomment-955536435

### DIFF
--- a/lib/gettext/compiler.ex
+++ b/lib/gettext/compiler.ex
@@ -530,7 +530,7 @@ defmodule Gettext.Compiler do
             unquote(singular_fun)(unquote(msgctxt), unquote(msgid), bindings)
           ) do
             require Gettext.Interpolation
-            Gettext.Interpolation.compile_interpolate(unquote(msgstr), bindings)
+            Gettext.Interpolation.compile_interpolate(:translation, unquote(msgstr), bindings)
           end
         end
     end
@@ -563,7 +563,12 @@ defmodule Gettext.Compiler do
           quote do
             unquote(form) ->
               require Gettext.Interpolation
-              Gettext.Interpolation.compile_interpolate(unquote(str), var!(bindings))
+
+              Gettext.Interpolation.compile_interpolate(
+                :plural_translation,
+                unquote(str),
+                var!(bindings)
+              )
           end
         end)
 
--- a/lib/gettext/compiler.ex
+++ b/lib/gettext/compiler.ex
@@ -2,7 +2,6 @@ defmodule Gettext.Compiler do
   @moduledoc false
 
   alias Gettext.{
-    Interpolation,
     PO,
     PO.Translation,
     PO.PluralTranslation
@@ -518,7 +517,6 @@ defmodule Gettext.Compiler do
     msgid = IO.iodata_to_binary(t.msgid)
     msgstr = IO.iodata_to_binary(t.msgstr)
     msgctxt = t.msgctxt && IO.iodata_to_binary(t.msgctxt)
-    keys = Interpolation.keys(msgstr)
 
     case msgstr do
       # Only actually generate this function clause if the msgstr is not empty.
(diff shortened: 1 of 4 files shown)


lib/gettext/interpolation.ex
```diff
@@ -148,4 +160,69 @@ defmodule Gettext.Interpolation do
...
+  This macro takes a static message and some dynamic bindings. The generated
+  code will return an `{:ok, interpolated_string}` tuple if the interpolation
+  is successful. If it encounters a binding in the message that is missing from
+  `bindings`, it returns `{:missing_bindings, incomplete_string, missing_bindings}`,
+  where `incomplete_string` is the string with only the present bindings interpolated
+  and `missing_bindings` is a list of atoms representing bindings that are in
+  `interpolatable` but not in `bindings`.
+  """
+  @doc since: "0.19.0"
+  defmacro compile_interpolate(message, bindings) do
```
maennchen, line 176: @josevalim This does not include the arg `:translation | :plural_translation`, since I could not identify what it actually does.

If it is still needed, wheat was the job of the argument?

lib/gettext/interpolation.ex
```diff
@@ -148,4 +159,68 @@ defmodule Gettext.Interpolation do
...
+  This macro takes a static message and some dynamic bindings. The generated
+  code will return an `{:ok, interpolated_string}` tuple if the interpolation
+  is successful. If it encounters a binding in the message that is missing from
+  `bindings`, it returns `{:missing_bindings, incomplete_string, missing_bindings}`,
+  where `incomplete_string` is the string with only the present bindings interpolated
+  and `missing_bindings` is a list of atoms representing bindings that are in
+  `interpolatable` but not in `bindings`.
+  """
+  defmacro compile_interpolate(message, bindings) do
+    unless is_binary(message) do
```
maennchen, line 175: I decided to add this check and force the usage of `runtime_interpolate/2`, if the message is not static.

This is to ensure that the compiler is actually able to optimize all calls properly.

lib/gettext/compiler.ex
```diff
@@ -572,7 +560,11 @@ defmodule Gettext.Compiler do
       # so when quoted they are a list.
       clauses =
         Enum.flat_map(msgstr, fn {form, str} ->
-          quote do: (unquote(form) -> unquote(compile_interpolation(str, :plural_translation)))
+          quote do
+            unquote(form) ->
+              require Gettext.Interpolation
```
josevalim, line 565: Tip: if you require it outside of the quote, it should be required inside the quote too (iirc the information is passed forward).

lib/gettext/compiler.ex
```diff
@@ -622,67 +614,6 @@ defmodule Gettext.Compiler do
...
-
-    all_bindings_clause = quote do: (unquote(match) -> {:ok, unquote(interpolation)})
-
-    dynamic_interpolation_clause =
-      quote do
-        %{} -> Gettext.Interpolation.interpolate(unquote(interpolatable), var!(bindings))
-      end
-
-    clauses =
-      if keys == [:count] and translation_type == :plural_translation do
```
josevalim, line 653: The `translation_type` is used here to avoid adding dynamic interpolation clauses if we know there are no bindings. So we need to add it back and add this clause back? :)

josevalim (commented):
This looks great! I have dropped only two minor comments, and this is ready to go. Once we merge this, then we can discuss how to add a callback and ship it!

lib/gettext/compiler.ex
maennchen: @josevalim Ah, I think I finally get it: If there's only a `count` parameter for a translation plural, it is impossible for count to be empty, since gettext itself will set the binding. Therefore only the static version makes sense.

lib/gettext/compiler.ex
maennchen: @josevalim I tried requiring both at the top of the module as well as in the function around the quote. In both cases I got the following error:

```
you must require Gettext.Interpolation before invoking the macro Gettext.Interpolation.compile_interpolate/3
```

I however think that this is not strictly necessary, since with #291 the module will have to be required globally anyways.

lib/gettext/compiler.ex
maennchen: Fixed with cf6b9902e06c6794a409365d75ee7914f0e0f2d1

josevalim:

Issue #294, opened 2021-12-01
Interpolation Behaviour Message Format


maennchen:
Right now, this library adds a flag `elixir-format` to each extracted message.

When using another interpolation format, that might be the wrong format. For ICU, it should be `icu-format` for example.

I, therefore, propose adding a new callback `message_format` to the `Gettext.Interpolation` module. This one would return a string for the resulting flag.

If you would prefer, that callback could also be optional and fallback to `elixir-format` by default. I think making it required and therefore requiring implementors to think about what is correct would be a better way though.

If you agree, I'll create a PR for this as well.

This request comes out of the discussion of an implementation in `ex_cldr_messages` here: https://github.com/elixir-cldr/cldr_messages/pull/3

maennchen:
Solved in #295

PR #295, opened 2021-12-01
Interpolation Behaviour Message Format

maennchen:
Implements #294

Merge #296 first

This is only a draft and not a final implementation.

It looks like it is expected to configure the extract via the `mix.exs` options and not via the backend. Why is that?

If we actually went this way, it would probably make more sense to yield the format when extracting. This way the extractor itself would not need to know about what format to use.


josevalim:
Hi @maennchen, thanks for the PR!

I believe the changes could be simpler. Change `flags: MapSet.new([extracted_translations_flag])` to be the flag we got from the interpolator and call it a day? :)

maennchen:
@josevalim Wouldn't that mean, that no messages would be purged if extracting with another interpolator?

(Calling `PO.Translations.autogenerated?`)

Is it ok to use the backend config in extraction? If yes: why are there configs via the `mix.exs` and not only via the backend module?

josevalim:
The configs in mix.exs are because we don't know what are the backends until that moment, iirc.

josevalim:
@maennchen should we use another flag for auto generated? What does gettext use?

maennchen:
@josevalim I'm not aware if gettext itself does something like that. It seems like there's no (documented) flag for that: https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html#index-msgstr

maennchen:
But I think we can just make one up as long as we prefix it. (To make sure it doesn't collide)

Something like `elixir-generated`.

If you would want to go down that path, I would also offer a PR and do that before this PR.

josevalim:
Let's go with elixir-generated, yeah. :+1:

lib/gettext/po/translations.ex
```diff
@@ -14,7 +14,7 @@ defmodule Gettext.PO.Translations do
   @doc """
   Tells whether a translation was manually entered or generated by Gettext.
 
-  As of now, a translation is considered autogenerated if it has the "elixir-format" flag.
+  As of now, a translation is considered autogenerated if it has the "elixir-autogenerated" flag.
```
maennchen, line 17: TODO: CHange when #296 is decided & merged.

josevalim:
Can you please rebase? And then I think this should be good!

maennchen:
@josevalim Done

josevalim:

PR #296, opened 2021-12-06
Separate Flag to detect autogenerated Translations

maennchen:
Follow up from https://github.com/elixir-gettext/gettext/pull/295#issuecomment-983994697


lib/gettext/po/translations.ex
```diff
@@ -29,7 +29,8 @@ defmodule Gettext.PO.Translations do
   """
   @spec autogenerated?(PO.translation()) :: boolean
   def autogenerated?(%module{flags: flags} = _translation) when is_translation(module) do
-    MapSet.member?(flags, "elixir-format")
+    # TODO: Deprecate elixir-format flag to detect autogenerated translations
+    not MapSet.disjoint?(flags, MapSet.new(["elixir-autogenerated", "elixir-format"]))
```
maennchen, line 33: How would you like to proceed with this?

- like it is now
- remove it
- show some kind of warning

josevalim: `MapSet.member?(flags, "elixir-format") or MapSet.member?(flags, "elixir-autogenerated")` would be my vote. We can leave a note to remove it in the future.

josevalim (commented):
Looks good to me, just one tiny comment.

Actually, I am thinking `elixir-autogenerated` is a bit too long and we may have thousands of translations. Maybe `elixir-gen` or `elixir-autogen`? Or `from-elixir`?

maennchen:
@josevalim i like the long name because it is very descriptive of what it is used for and storage is cheap. I like elixir-autogen most of your suggestions.

josevalim:
@whatyouhide what is your pick? :)

lib/gettext/po/translations.ex
maennchen: @josevalim Done

whatyouhide:
`ex-autogen` would be my pick, if we're going for shortness. Otherwise `elixir-gen` or `elixir-autogen` are both fine!

maennchen:
@josevalim / @whatyouhide I changed it to `elixir-autogen`.

maennchen:
@whatyouhide Should've read your comment more carefully. `ex-autogen` is also a good option. I'll change it.

lib/gettext/po/translations.ex
```diff
@@ -29,7 +29,8 @@ defmodule Gettext.PO.Translations do
   """
   @spec autogenerated?(PO.translation()) :: boolean
   def autogenerated?(%module{flags: flags} = _translation) when is_translation(module) do
-    MapSet.member?(flags, "elixir-format")
+    # TODO: Deprecate elixir-format flag to detect autogenerated translations
```
josevalim, line 32:
```suggestion
    # TODO: Remove elixir-format flag to detect autogenerated translations in future releases
```

josevalim:
Just one minor comment and we can ship it.

lib/gettext/po/translations.ex
maennchen: Done

josevalim:

PR #297, opened 2021-12-14
Supress Warning "variable "bindings" is unused"

maennchen:
Suppresses warning:

```
warning: variable "bindings" is unused (if the variable is not meant to be used, prefix it with an underscore)
  lib/hygeia_gettext.ex:1: HygeiaGettext.T_fr_default.fr_default_lngettext/5
```

This warning happens, because `clauses` can be an empty list. Therefore the variable is actually never used.

### DIFF
--- a/lib/gettext/compiler.ex
+++ b/lib/gettext/compiler.ex
@@ -611,7 +611,7 @@ defmodule Gettext.Compiler do
               line: unquote(t.po_source_line)
         end
 
-      quote do
+      quote generated: true do
         Kernel.unquote(kind)(
           unquote(plural_fun)(
             unquote(msgctxt),


maennchen:
The CI failure seems unrelated.

josevalim:
@maennchen do we want to raise if there are no clauses? or is this expected?

maennchen:
@josevalim As I understand it, this can happen if there is no translation for a specific msgstr (for the given locale/domain).

In this case, this is expected. The translations for some locales are done asynchronously and might take a bit longer.

I think the possibility of not having translated an entry should also be considered in general in the gettext library.

josevalim:

Issue #210, opened 2019-08-13
Keeping or deleting obsolete translations


whatyouhide:
When we are merging an old PO file with a new POT file, we remove all translations in the old PO that don't match anything in the new POT. However, the `msgmerge` tool actually marks them as obsolete:

```po
#~ msgid "hello"
#~ msgstr "ciao"
```

Should we do the same, or are we good with removing them?

\cc @josevalim

josevalim:
@whatyouhide sounds good to me!

halostatue:
It would also be good to consider the `changed` marker if it can be detected:

```
#| msgid "Old Translation Key"
msgid "New Translation Key"
msgstr ""
```

PR #298, opened 2021-12-19
Fix failing test

whatyouhide:
This was the failure:

```
  1) test extracting and extracting with --merge (Mix.Tasks.Gettext.ExtractTest)
     test/mix/tasks/gettext.extract_test.exs:19
     Assertion with == failed
     code:  assert read_file("priv/gettext/it/LC_MESSAGES/my_domain.po") == "#, elixir-format, ex-autogen\n#: lib/other.ex:3\nmsgid \"other\"\nmsgstr \"\"\n"
     left:  ""
     right: "#, elixir-format, ex-autogen\n#: lib/other.ex:3\nmsgid \"other\"\nmsgstr \"\"\n"
     stacktrace:
       test/mix/tasks/gettext.extract_test.exs:62: (test)
```

It was caused by the second file written in the test not being recompiled.

cc @josevalim do you know why changing the manifests in `gettext.extract` was not enough, and I had to use `--force`? Did something change in Elixir itself?

### DIFF
--- a/lib/mix/tasks/gettext.extract.ex
+++ b/lib/mix/tasks/gettext.extract.ex
@@ -47,8 +47,6 @@ defmodule Mix.Tasks.Gettext.Extract do
     mix_config = Mix.Project.config()
     {opts, _} = OptionParser.parse!(args, switches: @switches)
     pot_files = extract(mix_config[:app], mix_config[:gettext] || [])
-    IO.inspect(pot_files, label: "pot_files")
-    IO.inspect(opts, label: "opts")
 
     if opts[:check_up_to_date] do
       run_up_to_date_check(pot_files)
--- a/lib/mix/tasks/gettext.extract.ex
+++ b/lib/mix/tasks/gettext.extract.ex
@@ -47,6 +47,8 @@ defmodule Mix.Tasks.Gettext.Extract do
     mix_config = Mix.Project.config()
     {opts, _} = OptionParser.parse!(args, switches: @switches)
     pot_files = extract(mix_config[:app], mix_config[:gettext] || [])
+    IO.inspect(pot_files, label: "pot_files")
+    IO.inspect(opts, label: "opts")
 
     if opts[:check_up_to_date] do
       run_up_to_date_check(pot_files)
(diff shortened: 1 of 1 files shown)


josevalim:
@whatyouhide it is because the compiler got smarter and it is no longer mtime based.

Although your update means that Elixir code will be compiled twice... I would probably keep only `mix compile --force`. That should be enough.

whatyouhide:
@josevalim actually, with `mix compile --force` + `mix compile.elixir` the test is still failing. With `mix compile` + `mix compile.elixir --force` it seems to work 🤔 Pushed a commit, in any case.

josevalim:
That will likely do a double compilation if the user didn't compile the project before. I would check in an actual project... it may be a test bug exclusively.

whatyouhide:
@josevalim I don't think it's a test bug, because this is the code we are running

```elixir
    Enum.map(Mix.Tasks.Compile.Elixir.manifests(), &make_old_if_exists/1)
    Mix.Task.reenable("compile.elixir")
    Mix.Task.run("compile")
    Mix.Task.run("compile.elixir")
```

If we run `compile --force` but the user already ran `mix compile`, then we do not recompile Elixir files. At that point, we do run `compile.elixir` (because we reenabled it), but since the mtime of the manifest is not used anymore, we don't recompile anything either. Does that make sense?

Btw, what the test is doing is this:

  1. Write at `.pot` file
  2. Run `gettext.extract`
  3. Write another `.po` file
  4. Re-run `gettext.extract`, which does not recompile the Gettext backend for the reason above (so doesn't pick up anything)

josevalim:
You are correct, apologies. --force is independent of reenable and that was what confused me. We will need to release a new version too, i think it should be a minor one. Do you plan to do it too? If so, don't forget to update ExDoc. If you can't, let me know and I will do it!

lib/mix/tasks/gettext.extract.ex
```diff
@@ -47,6 +47,8 @@ defmodule Mix.Tasks.Gettext.Extract do
     mix_config = Mix.Project.config()
     {opts, _} = OptionParser.parse!(args, switches: @switches)
     pot_files = extract(mix_config[:app], mix_config[:gettext] || [])
+    IO.inspect(pot_files, label: "pot_files")
+    IO.inspect(opts, label: "opts")
```
josevalim, line 51:
```suggestion
```

whatyouhide: Lol yes I saw that I left this in there yesterday and still managed to forget taking it out 😌

whatyouhide:
@josevalim I'll fix #210 (I have a WIP) and release, sure!

PR #300, opened 2022-01-06
Fix @spec for Gettext.put_locale/{1, 2}

kipcole9:
Gettext.put_locale/{1,2} delegates to Process.put/2
which returns the previously set value for the key which
may be nil or term().

The @spec for Gettext.put_locale/{1,2}
is updated to specify the return of `locale | nil`
to reflect this.


josevalim:

PR #302, opened 2022-02-03
Merge elixir-autogen flag on extraction

### LINKED ISSUE #301: Can you explain `ex-autogen` consequences?
Hi,
I just noticed that translations automatically generated get flagged with `ex-autogen`. I read in the [doc](https://hexdocs.pm/gettext/Mix.Tasks.Gettext.Extract.html) that it allows to automatically remove a translation for the po files if it is not used anymore. Does it mean that the existing translations (created before the existence of the flag) need to be updated with this flag as well?
Thanks

maennchen:
Merge elixir-autogen flag on extraction
Remove old ex-autogen flag on extraction
Warn when using ex-autogen or only elixir-format flag

Fixes #301


lib/mix/tasks/gettext.extract.ex
```diff
@@ -18,6 +18,15 @@ defmodule Mix.Tasks.Gettext.Extract do
   If a translation from the POT is no longer present and has the `ex-autogen`
   flag, the translation will be removed.
 
+  Before `v0.19.0`, the `elixir-format` flag was used to detect automatically
+  extracted translations. This has been deprecated in `v0.19.0`. When extracing
```
josevalim, line 22:
```suggestion
  extracted translations. This has been deprecated in `v0.19.0`. When extracting
```

josevalim:
I have an idea: let's rename elixir-format to ex-format? This way elixir-format is completely phased out, which I think will be clearer than repurposing its meaning. Thoughts?

josevalim:
If we go with the suggestions above, I think the code can be simpler: we simply remove any elixir-format flag if we see it.

maennchen:
@josevalim I think changing the flag name is a breaking change that could be avoided. This flag is used for editors to recognize our `%{...}` placeholders and changing it could mean that we break some editor integrations.

josevalim:
It is a breaking change but at the same time now is the time to make it, before 1.0?

maennchen:
@josevalim If this was my choice alone, I would decide against changing it for the following reasons:

* Upside: We can remove the flag when we see it and add the autogen one => a bit clearer migration
* Downside 1: Breaking Change
* Downside 2: I think the name of the flag is more explicit and the chance is higher that developers of PO Editors find/add the format.

If you think the upside to this change is bigger than the downside, I'll be swayed to do the change as well. Just let me know what you think is the better way.

josevalim:
Yeah, so I think I messed up in pushing for `ex-autogen`. It should be `elixir-autogen` and `elixir-format` then, at least for consistency? Good news is that we didn't rewrite elixir-format yet, so there is time for this change?

maennchen:
@josevalim Sure :)

I'll append that change as well to this PR :)

maennchen:
@josevalim I pushed an update and also added some Logger warnings when using deprecated flags.

lib/gettext/po/translations.ex
```diff
@@ -28,9 +30,38 @@ defmodule Gettext.PO.Translations do
...
+
+        Translation:
+        #{msgid}
+        """)
+
+        true
+
+      true ->
+        false
+    end
```
josevalim, line 64: I would not worry about this for now. Maybe we can introduce it 2 or 3 versions down the road because they should have upgraded by then. We probably don't even need to check for ex-autogen because most people will still have elixir-format and the one person who might not is the one drafting this PR. :D

```suggestion
    MapSet.member?(flags, "elixir-autogen") or MapSet.member?(flags, "elixir-format")
```

maennchen: @josevalim I reverted to the simple condition without logging but included `ex-autogen`. Is that ok?

josevalim:

Issue #303, opened 2022-03-03
Support falling back to another language, if a PO for a language is partially translated


tusooa:
Currently the default implementation fallback to msgid if the translation for a language is missing. However, in many cases, we would want the user to choose a fallback language, so that if there is some missing translation, they can still read in what they would prefer.

There is handle_missing_translation, but a big problem is that it does not get msgctxt, so if we would like to try gettext again with some other language, we do not have enough information to do so -- suppose the initial call is `dpgettext` or so.

A simple solution would be pass msgctxt to handle_missing_translation functions. Or just support fallbacking natively in the module.

## Downstream
https://git.pleroma.social/pleroma/pleroma/-/merge_requests/3634#note_89751

josevalim:
A pull request is welcome. We should define a new version of those callbacks that receive the context. Later I can do a quick pass to make it backwards compatible.

whatyouhide:
@tusooa any updates on this?

tusooa:
> @tusooa any updates on this?

We already implemented support for fallbacks in Pleroma.

whatyouhide:
@tusooa fantastic 🙃 Do you think that's something that we should backport back to Gettext, or were you able to do this with the current Gettext API, meaning no changes are needed to Gettext?

whatyouhide:
Okay, so for now I'll close this then. Thanks for reporting back! 💟

PR #304, opened 2022-03-03
Pass msgctxt into handling missing translation functions

tusooa:
To resolve msgctxt not being passed into the fallback function,
we added new callbacks handle_missing_translation/5 and
handle_missing_plural_translation/7 that will get msgctxt.
The default implementations for these two functions call the
old versions respectively, so users implementing the old callbacks
still get the same result.

BUG: https://github.com/elixir-gettext/gettext/issues/303

### DIFF
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -631,6 +631,22 @@ defmodule Gettext do
       end
 
       defoverridable handle_missing_translation: 4, handle_missing_plural_translation: 6
+
+      def handle_missing_translation(locale, domain, msgctxt, msgid, bindings),
+        do: handle_missing_translation(locale, domain, msgid, bindings)
+
+      def handle_missing_plural_translation(
+            locale,
+            domain,
+            msgctxt,
+            msgid,
+            msgid_plural,
+            n,
+            bindings
+          ),
+          do: handle_missing_plural_translation(locale, domain, msgid, msgid_plural, n, bindings)
+
+      defoverridable handle_missing_translation: 5, handle_missing_plural_translation: 7
     end
   end
 
--- a/lib/gettext/backend.ex
+++ b/lib/gettext/backend.ex
@@ -70,6 +70,25 @@ defmodule Gettext.Backend do
   @callback handle_missing_translation(
               Gettext.locale(),
               domain :: String.t(),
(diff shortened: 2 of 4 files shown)


whatyouhide (approved):
Nice, I love it!

lib/gettext.ex
```diff
@@ -631,6 +631,22 @@ defmodule Gettext do
...
+            domain,
+            msgctxt,
+            msgid,
+            msgid_plural,
+            n,
+            bindings
+          ),
+          do: handle_missing_plural_translation(locale, domain, msgid, msgid_plural, n, bindings)
+
+      defoverridable handle_missing_translation: 5, handle_missing_plural_translation: 7
```
josevalim, line 649: This looks great! The only thing missing is a `@before_compile` hook that uses `Module.defines?` to see if the user is defining the old versions and emit a deprecation warning. :)

lib/gettext/backend.ex
```diff
@@ -78,9 +97,30 @@ defmodule Gettext.Backend do
...
+  Fallback handling for missing translations.
+
+  Same as `c:handle_missing_plural_translation/7`, but without msgctxt. This is kept
+  for compatibility with older versions of this library.
+
+  The default implementation of `c:handle_missing_plural_translation/7` will call this
+  function, so users implementing the 6-argument callback will still get the same
+  results. However, they are encouraged to switch to the new 7-argument version.
+  """
   @callback handle_missing_plural_translation(
```
josevalim, line 124: We can remove the old callbacks, let's keep only the new ones!

lib/gettext.ex
tusooa: I'm not sure how to do this. I tried to put
```
      if Module.defines?(env.module, {:handle_missing_translation, 4}) do
        IO.warn(
          "`handle_missing_translation/4` is deprecated; use `handle_missing_translation/5` instead",
          Macro.Env.stacktrace(env)
        )
      end
```
in Gettext.Compiler.__before_compile__, but it emits warnings for every module, even if it does not manually define this.

josevalim:
I will investigate it then, thank you!

josevalim:

Issue #215, opened 2019-09-06
Separate PO and Gettext into separate packages


kipcole9:
I am working on an ICU Message formatter and localisation processor as the (maybe) last part of the [ex_cldr](https://github.com/elixir-cldr/cldr). One requirement is of course to store and manage translations and PO files are clearly the defacto globally standard.

I would like to propose splitting Gettext into as PO package that looks after storage, parsing, merging and extracting.  That way different localisation packages can leverage the same infrastructure - which is supported by the `format` tag in a PO file. As you know, today Gettext saves all messages with the `format-elixir` tag.

What would need to change:

1. API and usage for current Gettext users would need to remain identical for backwards compatibility

2. The extractor would need to be told hold to tag messages so they aren’t all tagged as `format-elixir`

3. The backend interface would be normalised. Instead of `BackendModule.__gettext__(:priv)` for example, it would be better to have `BackendModule.__po__(:priv)` in order to be more specific and not tied to the Gettext API alone

4 Compiling PO files would compile messages based upon the message format.  Perhaps a behaviour would be introduced so that any message format could be compiled.  Instead of `use Gettext, ....` it would become `use PO, icu_format: MyMessageFormatter, elxiir_format: Gettext` or some such.  Then `use Gettext, ...` would also include `use PO, ....`

5 Lastly, this may open up the opportunity to have different storage models, like DB, or `:persistent_term` or even a web service that can replace PO *files* while still retaining the well known format.

I’m fine to work on this - its better than creating another PO file parser/merger/extracted as @tmbb as already had to do,

Thoughts?

tmbb:
> I’m fine to work on this - its better than creating another PO file parser/merger/extracted as @tmbb as already had to do

I actually think Gettext could use a new .po parser

josevalim:
Agreed on a new parser but that is just part of the work. We can probably re-use everything else. I think Andrea considered a `nimble_parsec`-based PO parser but I am not sure of the status. Contributions are definitely welcome!

I am also completely happy to go in the direction proposed here. My suggestion is to start adding the new API directly to this repository. We will introduce new modules, move the API over, and once we are happy with the results, we extract it to a separate project. By then, I think we should also create an elixir-gettext org but no need to worry about it for now.

What do you think?

tmbb:
I believe I have a functional .po parser based on nimble_parsec already. I
have to look for it.

kipcole9:
[...]

PR #306, opened 2022-04-12
Replace PluralForm implementation with Expo library

maennchen:
I would prefer to open a PR per feature that is externalized.

Would it make sense to create a separate branch than `main` that collects all the features and I point my PRs to so that we can review them separately and still merge them in bulk?

Partial implementation of #215


whatyouhide:
Yep, this one is pretty straightforward, let's give it a try. I’m okay to make these PRs against `main` as long as we don't release Gettext from `main` until the transition is finished. I don't want to put new transient dependencies in people's projects until we're sure the move will be complete.

PR #307, opened 2022-04-14
Replace PO parse / compose with Expo library

maennchen:
Partial implementation of #215

Blocked by elixir-gettext/expo#30

This is only an initial draft and not the finished implementation. I'll be on vacation until 25th and will probably not work on this until then.


maennchen:
@whatyouhide I would really like to finish up this PR. Will you be able to have a look at the outstanding parser issue or should I try to find some help somewhere else?

whatyouhide:
@maennchen hey, sorry for the wait. I was spending my OSS cycles mostly on [Xandra](https://github.com/lexhide/xandra) the past few weeks, and I have just a little more to go. I would be able to take a look at this in something like a couple of weeks, would that works? If I'll get to it sooner, I'll get to it sooner 😄

maennchen:
@whatyouhide Thanks for the feedback. Xandra sounds interesting, I'll have to look at that one a bit closer :)

I think I'm going to attempt it once more on my own in the meantime and if that doesn't work I'll wait for you.

maennchen:
@whatyouhide I tried in a new way and I nearly got it working.

To get to this point, I added a `$end` token and tried to get rid of as many `$empty` states as possible.

There are now only 2 errors left. It seems like the `previous` and `obsolete` tokens are somehow consumed prematurely.

Details: https://github.com/elixir-gettext/expo/pull/30

maennchen:
@josevalim FYI: I quickly rebased & fixed the tests.

I will have to do one more pass over the PR and make sure all the naming etc. is correct.

maennchen:
@josevalim / @whatyouhide I've completed the integration.

A few points to note:

* Look at the commits separately
  - commit 1 is the actual integration
  - commit 2 renames translations to messages (except public function names etc.; as discussed with @josevalim via slack)
  - commit 2 has no actual code changes, it's all just names
* Even though `expo` already supports `previous` / `obsolete`, that has no noticeable impact to the user of `gettext`
  - previous translations are kept on merge. no new ones are added
  - obsolete translations are discarded on compile - 
  - obsolete translations are kept on merge. no new ones are added
  - since neither were supported and would've been broken by merge previously, I think it's a safe assumption that our users do not have any of those anyways
 * flags `MapSet` => nested lists
   - needed so that the file formatting does not change when merging po files (order & line separation)

Thank you both for your time :)

lib/gettext/compiler.ex
```diff
@@ -595,26 +597,26 @@ defmodule Gettext.Compiler do
...
          interpolation_module
        ) do
-    warn_if_missing_plural_forms(locale, plural_mod, t, file)
+    warn_if_missing_plural_forms(locale, plural_mod, message, file)
 
-    msgid = IO.iodata_to_binary(t.msgid)
-    msgid_plural = IO.iodata_to_binary(t.msgid_plural)
-    msgstr = Enum.map(t.msgstr, fn {form, str} -> {form, IO.iodata_to_binary(str)} end)
-    msgctxt = t.msgctxt && IO.iodata_to_binary(t.msgctxt)
+    msgid = IO.iodata_to_binary(message.msgid)
```
josevalim, line 612: Do we still need to keep the iodata_to_binary cals? Aren't those normalized already?

lib/gettext/compiler.ex
```diff
@@ -558,23 +560,23 @@ defmodule Gettext.Compiler do
...
          singular_fun,
          _plural_fun,
          _file,
          _plural_mod,
          interpolation_module
        ) do
-    msgid = IO.iodata_to_binary(t.msgid)
-    msgstr = IO.iodata_to_binary(t.msgstr)
-    msgctxt = t.msgctxt && IO.iodata_to_binary(t.msgctxt)
+    msgid = IO.iodata_to_binary(message.msgid)
```
josevalim, line 573: Same here, do we need to keep the iodata_to_binary calls?

lib/gettext/extractor_agent.ex
```diff
@@ -34,16 +36,16 @@ defmodule Gettext.ExtractorAgent do
...
+    key = Message.key(message)
 
     Agent.cast(@name, fn state ->
       # Initialize the given backend to an empty map if it wasn't there.
-      state = update_in(state.translations, &Map.put_new(&1, backend, %{}))
+      state = update_in(state.messages, &Map.put_new(&1, backend, %{}))
 
-      update_in(state, [:translations, backend, domain], fn translations ->
-        Map.update(translations || %{}, key, translation, &merge_translations(&1, translation))
+      update_in(state, [:messages, backend, domain], fn messages ->
```
josevalim, line 47:
```suggestion
      update_in(state.messages[backend][domain], fn messages ->
```

lib/gettext/merger.ex
```diff
@@ -1,13 +1,13 @@
...
-  alias Gettext.PO.Translation
-  alias Gettext.PO.PluralTranslation
+  alias Expo.Po
+  alias Expo.Message
+  alias Expo.Messages
   alias Gettext.Fuzzy
 
   @new_po_informative_comment """
-  ## "msgid"s in this file come from POT (.pot) files.
+  # "msgid"s in this file come from POT (.pot) files.
```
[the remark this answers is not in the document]
maennchen: @josevalim In `expo`, the `#` is automatically prepended since otherwise it would be possible to write a comment without the `#` which then would not actually be a comment.

Therefore I had to remove it here.

lib/gettext/compiler.ex
maennchen: @josevalim A msgid from `expo` is a list of strings to preserve the lines. Therefore it is still needed.

To format the contents of the `.po` nicely, `Expo.message.[Singular|Plural].rebalance` is provided which will put every line of text onto a new line in the `.po` as well.

test/gettext/merger_test.exs
```diff
@@ -1,139 +1,176 @@
-defmodule Gettext.MergerTest do
+defmodule GettextMergerTest do
```
josevalim, line 1:
```suggestion
defmodule Gettext.MergerTest do
```

lib/gettext/extractor_agent.ex
maennchen: Done in abca45d686556fcec0dd010c97735be1bae83fc5

josevalim:

maennchen:
@josevalim Oh wow, I was not expecting for this to get through that easy. Thanks :heart:

josevalim:
We just need an expo release and we can start working on the other issues. :)

Issue #308, opened 2022-04-26
Uniq comments from gettext_comment/1 for the same translation (on extraction)


gabrielgiordan:
Hello! First, thanks to all contributors, the lib is awesome.

So, I'm using gettext inside an umbrella project, where 2 apps are using the same gettext module.

Inside the project, each app uses the same translation repeatedly in different places, for example, a call for `dgettext("errors", "App is outdated")` on app 1 and app 2.

I would like to add only one comment for this translation, but `gettext_comment/1` adds a new comment for each of its call.

How should I deal with it? Should I put `gettext_comment/1` on some kind of central place, perhaps inside the own gettext module?

Something like `dgettext_comment("some comment", "errors", "App is outdated")` would be helpful in this case

I tried to add the comment `#.` directly on the `.pot` file as well, but the `mix gettext.extract` task overrides it.

whatyouhide:
Hey @gabrielgiordan, good catch and thanks for the report! 💟 

We should support this at the library level. When doing the extraction, we should `uniq` the comments for a given translation.

Wanna take a stab at a PR? 🙃

gabrielgiordan:
Sure, let me take a look 🙂

whatyouhide:
Closed via #309.

PR #309, opened 2022-05-02
Uniq comments for same translation

gabrielgiordan:
Solves https://github.com/elixir-gettext/gettext/issues/308

Distinct comments for a same Gettext text on extraction, so a `gettext_comment/1` can be added into multiple places inside one or more apps, without duplications at the final `.pot` file.

### DIFF
--- a/lib/gettext/extractor_agent.ex
+++ b/lib/gettext/extractor_agent.ex
@@ -75,6 +75,6 @@ defmodule Gettext.ExtractorAgent do
   defp merge_translations(t1, t2) do
     t1
     |> Map.put(:references, t1.references ++ t2.references)
-    |> Map.put(:extracted_comments, t1.extracted_comments ++ t2.extracted_comments)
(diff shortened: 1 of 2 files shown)


josevalim (approved):

lib/gettext/extractor_agent.ex
```diff
@@ -75,6 +75,6 @@ defmodule Gettext.ExtractorAgent do
   defp merge_translations(t1, t2) do
     t1
     |> Map.put(:references, t1.references ++ t2.references)
-    |> Map.put(:extracted_comments, t1.extracted_comments ++ t2.extracted_comments)
+    |> Map.put(:extracted_comments, Enum.uniq(t1.extracted_comments ++ t2.extracted_comments))
```
whatyouhide, line 78: This is great, I think it actually solves a broader scope of comment-related issues 😄

PR #313, opened 2022-07-20
Bring back previous gettext plural implementation

josevalim:
It is simpler and requires considerably less infrastructure.


josevalim:

PR #315, opened 2022-07-20
Support marking messages as obsolete

maennchen:
Implements #210


lib/gettext.ex
```diff
@@ -524,6 +524,9 @@ defmodule Gettext do
       messages are appended to the file. If `:sort_by_msgid` is set to `true`,
       existing and new messages will be mixed and sorted alphabetically by msgid.
 
+    * `:mark_as_obsolete` - a boolean that indicates if obsolete messages should
+      be deleted or marked as obsolete. By default messages are deleted.
+
```
whatyouhide, line 529: I think it might be clearer to have this option as `:on_obsolete`:

```suggestion
    * `:on_obsolete` - controls what happens when obsolete messages are found.
      If `:mark_as_obsolete`, messages are kept and marked as obsolete.
      If `:delete`, obsolete messages are deleted. Defaults to `:delete`.

```

maennchen: @whatyouhide Agreed, I will adapt it accordingly next week.

lib/mix/tasks/gettext.merge.ex
```diff
@@ -109,7 +112,8 @@ defmodule Mix.Tasks.Gettext.Merge do
     locale: :string,
     fuzzy: :boolean,
     fuzzy_threshold: :float,
-    plural_forms: :integer
+    plural_forms: :integer,
+    mark_as_obsolete: :boolean
```
maennchen, line 116: @whatyouhide How should I handle this one?

* Keep as boolean
* Use string `delete` / `mark_as_obsolete`

If using a string, what do I do if the string is neither option?

lib/gettext.ex
```diff
@@ -524,6 +524,10 @@ defmodule Gettext do
       messages are appended to the file. If `:sort_by_msgid` is set to `true`,
       existing and new messages will be mixed and sorted alphabetically by msgid.
 
+    * `:on_obsolete` - controls what happens when obsolete messages are found.\
```
whatyouhide, line 527: We don't need the backslash here. Markdown won't render the newline anyways.

lib/mix/tasks/gettext.merge.ex
whatyouhide: I would go with `--on-obsolete=delete` or `--on-obsolete=mark_as_obsolete`. If it's neither, we `Mix.raise()` an error accordingly. Defaults to `delete`. Thoughts?

lib/gettext.ex
maennchen: @whatyouhide I tested building the docs and the backslash works for me.

whatyouhide: I mean, if you don't use the backslash, Markdown still won't render a newline. So the backslash is not necessary here 🙃

maennchen: **without**

before

**with**

after

lib/mix/tasks/gettext.merge.ex
maennchen: @whatyouhide Done

lib/gettext.ex
maennchen: It's not working in the CLI, but it's also not making it worse:

image

whatyouhide: We don't need the backslashes. Let's remove them.

whatyouhide (approved):

lib/gettext.ex
maennchen: @whatyouhide Removed.

PR #316, opened 2022-07-20
Save previous message for fuzzy matches

maennchen:
Should this be the default behavior or hidden behind a flag?

### DIFF
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -606,6 +606,10 @@ defmodule Gettext do
       If `:mark_as_obsolete`, messages are kept and marked as obsolete.
       If `:delete`, obsolete messages are deleted. Defaults to `:delete`.
 
+    * `:store_previous_message_on_fuzzy_match` - a boolean that controls
+      wether to store the previous message text in case of a fuzzy match.
+      Defaults to `false`.
+
   """
 
   defmodule Error do
--- a/lib/gettext/merger.ex
+++ b/lib/gettext/merger.ex
@@ -90,6 +90,16 @@ defmodule Gettext.Merger do
               {:matched, match, fuzzy_merged} ->
                 stats_acc = update_in(stats_acc.fuzzy_matches, &(&1 + 1))
                 unused = Map.delete(unused, Message.key(match))
+
+                fuzzy_merged =
+                  if Keyword.get(opts, :store_previous_message_on_fuzzy_match, false) do
+                    Map.update!(fuzzy_merged, :previous_messages, fn previous ->
+                      Enum.uniq_by(previous ++ [match], &Message.key/1)
+                    end)
+                  else
+                    fuzzy_merged
+                  end
+
                 {fuzzy_merged, {stats_acc, unused}}
 
               :nomatch ->
--- a/lib/mix/tasks/gettext.merge.ex
+++ b/lib/mix/tasks/gettext.merge.ex
@@ -102,6 +102,9 @@ defmodule Mix.Tasks.Gettext.Merge do
       If `mark_as_obsolete`, messages are kept and marked as obsolete.
       If `delete`, obsolete messages are deleted. Defaults to `delete`.
 
+    * `--store-previous-message-on-fuzzy-match` - controls if the previous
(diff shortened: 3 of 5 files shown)


whatyouhide:
How does it affect what gets written to the PO files?

maennchen:
@whatyouhide When we fuzzy match on a message, the new message is in the msgid and the old one will be stored as a previous message.

It will look something like this:
https://github.com/elixir-gettext/expo/blob/d3b6377edf24884f290efcaae1f5e346c39e3555/test/expo/po_test.exs#L500

maennchen:
@whatyouhide It would probably make sense to handle this the same as #315

maennchen:
@whatyouhide Do you have an update so that I can finish their PR?

whatyouhide:
@maennchen sorry for the long wait. I agree, having this behind an option (like `:on_obsolete`) sounds great. By the way, `#|` is the "official" GNU Gettext comment for old fuzzy matches, right?

maennchen:
@whatyouhide I‘ll do that 👍 

No, the official way fur fuzzy is the fuzzy flag.

The pipe comment signals the previous mesage.

If I extrat the string Hello World and change it to Hello Worlds, we identify it as fuzzy. We no longer have the initial extracted string though.

With bigger translations, it can be helpful for the translator to know how the message changed so that he can apply the same change to the current translation.

whatyouhide:
Yeah I’m down for that 👍 My question was: do "previous translations" have an official comment in GNU Gettext?

maennchen:
@whatyouhide Yes, the `#|` syntax is official.

See: https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html

> ```
> white-space
> #  translator-comments
> #. extracted-comments
> #: reference…
> #, flag…
> #| msgid previous-untranslated-string
> msgid untranslated-string
> msgstr translated-string
> ```
> [...]
>
> The previous-untranslated-string is optionally inserted by the msgmerge program, at the same time when it marks a message fuzzy. It helps the translator to see which changes were done by the developers on the untranslated-string.

whatyouhide:
Ah that's fantastic. Let's go for the option then, and we'll be good to go here.

maennchen:
@whatyouhide I'll add that in next week :+1:

lib/gettext.ex
```diff
@@ -606,6 +606,11 @@ defmodule Gettext do
       If `:mark_as_obsolete`, messages are kept and marked as obsolete.
       If `:delete`, obsolete messages are deleted. Defaults to `:delete`.
 
+    * `:on_fuzzy_match` - controls what happens when fuzzy messages matches are found.
```
maennchen, line 609: @whatyouhide I'm not happy with the naming of this option. Do you have a better idea than this?

whatyouhide: @maennchen since the behaviours here are not complementary, but rather one a subset of the other, I think we can have `store_previous_message_on_fuzzy_match: boolean()`. The option name is quite long, but I like that it really conveys what it's doing. Regardless of whether the option is true or false, we'll still override the message and mark it as fuzzy, so the option only controls whether we store the old message too. Thoughts?

maennchen: @whatyouhide I agree that a longer more descriptive name is the way to go. I'll adapt accordingly.

maennchen: @whatyouhide Done, better?

lib/gettext.ex
```diff
@@ -606,6 +606,10 @@ defmodule Gettext do
       If `:mark_as_obsolete`, messages are kept and marked as obsolete.
       If `:delete`, obsolete messages are deleted. Defaults to `:delete`.
 
+    * `:store_previous_message_on_fuzzy_match` - a boolean that controls
+      wether to store the previous message text in case of a fuzzy match.
```
whatyouhide, line 610:
```suggestion
      whether to store the previous message text in case of a fuzzy match.
```

whatyouhide (approved):

whatyouhide:
Awesome work, thanks @maennchen! 💟

PR #322, opened 2022-10-07
Add `write_reference_line_numbers` option

### LINKED ISSUE #321: Possibility to remove line-numbers from extraction metadata
Would it be reasonable to add an option to exclude line numbers from the file references?

I totally love the references and their ability to communicate in which file a certain translation is used
but in most cases the file number isn't super relevant and is creating the need to re-extract the translations
even if I'm just making a newline in one of my files. It would also prevent small commits from creating a lot of changes.

I'd gladly help out by sending a PR but wanted to check in to see if it's something that would be valuable before committing my time? :)

johantell:
Background:
Since even the smallest changes in files can cause a cascade of changed line numbers in po/pot files we're often seeing larger amounts of changes in commits causing the need to re-extract than needed.

By adding the `write_reference_line_numbers` option developers has a way to decide for themselves whether they want to include them or not.

Fixes #321

### DIFF
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -521,6 +521,7 @@ defmodule Gettext do
 
     * `:write_reference_line_numbers` - a boolean that specifies whether file 
       reference comments include line numbers when outputting PO(T) files.
+      Defaults to `true`.
 
     * `:sort_by_msgid` - a boolean that modifies the sorting behavior.
       By default, the order of existing messages in a POT file is kept and new
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -519,6 +519,9 @@ defmodule Gettext do
       reference comments will not be written when extracting messages or merging
       messages, and the ones already found in files will be discarded.
 
+    * `:write_reference_line_numbers` - a boolean that specifies whether file 
+      reference comments include line numbers when outputting PO(T) files.
+
     * `:sort_by_msgid` - a boolean that modifies the sorting behavior.
       By default, the order of existing messages in a POT file is kept and new
       messages are appended to the file. If `:sort_by_msgid` is set to `true`,
--- a/lib/gettext/extractor.ex
+++ b/lib/gettext/extractor.ex
@@ -258,7 +258,7 @@ defmodule Gettext.Extractor do
        [
          new_pot_comment(),
          new_po
-         |> Merger.maybe_remove_references(gettext_config[:write_reference_comments])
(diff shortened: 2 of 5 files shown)


josevalim (approved):

lib/gettext/extractor.ex
```diff
@@ -258,7 +258,7 @@ defmodule Gettext.Extractor do
        [
          new_pot_comment(),
          new_po
-         |> Merger.maybe_remove_references(gettext_config[:write_reference_comments])
+         |> Merger.prune_references(gettext_config)
```
johantell, line 261: Would you sat that it would be beneficial to be explicit about what we're passing on here? like:

```elixir
> Merger.prune_references(Keyword.take(gettext_config, ~w[write_reference_comments write_reference_line_numbers]a))
```

lib/gettext/merger.ex
```diff
@@ -232,16 +232,38 @@ defmodule Gettext.Merger do
...
+        all
+    end
+  end
+
+  defp remove_references(%{references: _} = message), do: %{message | references: []}
 
-  def maybe_remove_references(%Messages{messages: messages} = all, true) do
-    messages = Enum.map(messages, &%{&1 | references: []})
-    %Messages{all | messages: messages}
+  defp remove_reference_line_numbers(%{references: references} = message) do
```
maennchen, line 256: references is a nested list:

* level 1: lines of multiple references 
* level 2: line with n references

type: https://hexdocs.pm/expo/Expo.Message.Singular.html#t:t/0

I therefore believe that you have to map twice

johantell: Yeah, perhaps it is better to be future safe than naively match on the list itself down in `remove_reference_line_number/1`. Would you be happy with this? e55e633

maennchen (approved):

lib/gettext.ex
```diff
@@ -519,6 +519,9 @@ defmodule Gettext do
       reference comments will not be written when extracting messages or merging
       messages, and the ones already found in files will be discarded.
 
+    * `:write_reference_line_numbers` - a boolean that specifies whether file 
+      reference comments include line numbers when outputting PO(T) files.
+
```
szymon-jez, line 524:
```suggestion
    * `:write_reference_line_numbers` - a boolean that specifies whether file 
      reference comments include line numbers when outputting PO(T) files.
      Defaults to `true`.

```

josevalim:

PR #326, opened 2022-11-15
Add `sort_by_msgid_case_insensitive`

### LINKED ISSUE #325: Sort messages alphabetically case insensitive
## Problem

Hi!

In my company we use an external tool to do translations. This tool sorts the messages by msgid alphabetically. The problem is that we use keys instead of English messages as msgid. Those keys include uppercase characters.

When extracting messages and `sort_by_msgid` flag to true, the keys are sorted alphabetically but the ones in uppercase are placed before the others, which causes us a lot of headaches.

## Proposed solution

Add another flag `sort_by_msgid_case_insensitive` to use a case insensitive comparation when extracting keys.

## PR

#326

superruzafa:
Closes #325.

### DIFF
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -597,10 +597,14 @@ defmodule Gettext do
       reference comments include line numbers when outputting PO(T) files.
       Defaults to `true`.
 
-    * `:sort_by_msgid` - a boolean that modifies the sorting behavior.
-      By default, the order of existing messages in a POT file is kept and new
-      messages are appended to the file. If `:sort_by_msgid` is set to `true`,
+    * `:sort_by_msgid` - modifies the sorting behavior. Can be either `nil` (the default),
+      `:case_sensitive`, or `:case_insensitive`.
+      By default or if `nil`, the order of existing messages in a POT file is kept and new
+      messages are appended to the file. If `:sort_by_msgid` is set to `:case_sensitive`,
       existing and new messages will be mixed and sorted alphabetically by msgid.
+      If set to `:case_insensitive`, the same applies but the sorting is case insensitive.
+      *Note*: this option also supports `true` and `false` for backwards compatibility,
+      but these values are deprecated as of v0.21.0.
 
     * `:on_obsolete` - controls what happens when obsolete messages are found.
       If `:mark_as_obsolete`, messages are kept and marked as obsolete.
--- a/lib/gettext/extractor.ex
+++ b/lib/gettext/extractor.ex
@@ -317,11 +317,30 @@ defmodule Gettext.Extractor do
 
     messages = old_and_merged ++ unique_new
 
+    sort_by_msgid =
+      case gettext_config[:sort_by_msgid] do
+        val when val in [:case_sensitive, :case_insensitive, nil] ->
+          val
+
+        bool when is_boolean(bool) ->
+          IO.warn("""
+          Passing a boolean to the :sort_by_msgid option is deprecated. \
+          Use :case_sensitive or :case_insensitive instead.\
+          """)
+
+          if bool, do: :case_sensitive, else: nil
+      end
+
     messages =
-      if gettext_config[:sort_by_msgid] do
-        Enum.sort_by(messages, &IO.chardata_to_string(&1.msgid))
-      else
-        messages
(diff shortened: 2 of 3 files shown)


lib/gettext.ex
```diff
@@ -528,6 +528,9 @@ defmodule Gettext do
       messages are appended to the file. If `:sort_by_msgid` is set to `true`,
       existing and new messages will be mixed and sorted alphabetically by msgid.
 
+    * `:sort_by_msgid_case_insensitive` - same as `:sort_by_msgid` but it
+      performs a case insensitive comparation when sorting alphabetically by msgid.
```
whatyouhide, line 532:
```suggestion
    * `:sort_by_msgid_case_insensitive` - a boolean that behaves similarly to 
      `:sort_by_msgid`, but that sorts with a case-insensitive comparison.
```

lib/gettext/extractor.ex
```diff
@@ -318,10 +318,15 @@ defmodule Gettext.Extractor do
...
-        messages
+      cond do
+        gettext_config[:sort_by_msgid] ->
+          Enum.sort_by(messages, & &1.msgid)
+
+        gettext_config[:sort_by_msgid_case_insensitive] ->
+          Enum.sort_by(messages, fn m -> Enum.map(m.msgid, &String.downcase/1) end)
+
+        true ->
+          messages
```
whatyouhide, line 329: I think we should think about this API, because as it stands in this PR you could pass `sort_by_msgid: true, sort_by_msgid_case_insensitive: false` and the sorting would work, even if by reading the options I wouldn't really know what to expect.

What do you think about going with two *complementary* options, such as the existing `:sort_by_msgid` and `:sort_is_case_sensitive`? You'd have to specify `sort_by_msgid: true, sort_is_case_sensitive: false` in your use case.

whatyouhide (commented):
Nice work! Left just a comment about the option name, but I think this feature in general is welcome 💯

lib/gettext/extractor.ex
whatyouhide: @josevalim thoughts?

lib/gettext/extractor.ex
```diff
@@ -318,10 +318,15 @@ defmodule Gettext.Extractor do
...
       end
```
maennchen, line 330: To make sure that the line splitting of the message has no impact on the sorting, I would recommend changing this a bit:

```suggestion
      cond do
        gettext_config[:sort_by_msgid] ->
          Enum.sort_by(messages, &IO.chardata_to_string(&1.msgid))

        gettext_config[:sort_by_msgid_case_insensitive] ->
          Enum.sort_by(messages, &(&1 |> IO.chardata_to_string() |> String.downcase()))

        true ->
          messages
      end
```

lib/gettext/extractor.ex
superruzafa: That's would be fine. In case you agree with that we could apply these changes in this PR.

lib/gettext/extractor.ex
whatyouhide: We were already sorting without `IO.chardata_to_string/1`. Do we know if it has ever caused issues? I think in general we should fix that in another PR and add test cases for it.

maennchen: @whatyouhide That’s something I should’ve probably done in #307 and spotted now when reading the changes.

I‘m not aware of issues so far but I can see how it coupd be an issue.

I can open a PR and add some tests if you prefer.

whatyouhide: Yeah, let's do that @maennchen, thanks 👍

maennchen: Done: https://github.com/elixir-gettext/gettext/pull/331

This PR will need a merge after #331 is merged.

whatyouhide: Merged #331 👍

whatyouhide:
@superruzafa I took the liberty of pushing a change to this branch that sets the `:sort_by_msgid` option to `:case_sensitive | :case_insensitive | nil`. @josevalim let me know if this looks good to you 🙃

maennchen (approved):

PR #327, opened 2022-12-16
Add a 'Basic Overview' docs section

nathanl:
I've been meaning to wrap my head around Gettext for a while. When I first looked at it, I was confused about how the English string could be the message id without that being very brittle. Now that I understand better, I wrote this introduction which (among other things) explains that.

### DIFF
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -4,15 +4,87 @@ defmodule Gettext do
   [gettext](https://www.gnu.org/software/gettext/)-based API for working with
   internationalized applications.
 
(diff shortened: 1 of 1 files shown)


josevalim (approved):

whatyouhide:
Love this, thanks @nathanl! 💟

nathanl:
@whatyouhide I also just started a repo for community translations of common strings (forgive me if this already exists somewhere): https://github.com/nathanl/elixir_common_i18n/tree/main/gettext

PR #331, opened 2022-12-21
Sort Messages independent of message line splitting

maennchen:


whatyouhide (approved):
Nice, thank you @maennchen! 💟

PR #335, opened 2023-01-09
Fix up-to-date check in extractor

### LINKED ISSUE #334: check-up-to-date fails even though files are just extracted
I was just about to try the new version (0.21.0) and noticed that the pipeline is failing:

```
** (Mix) mix gettext.extract failed due to --check-up-to-date.
The following POT files were not extracted or are out of date:
```

I tried deleting all the POT files and extracting them again, with the same result.

My gettext configuration looked like:

```elixir
    [
      write_reference_comments: false,
      sort_by_msgid: true
    ]
```
Because of the `sort_by_msgid` deprecation I tried `false`, `:case_insensitive` and `:case_sensitive` – same result.

Am I missing something on my side?

maennchen:
I can't come up with a small test-case. Without it the check fails in one of my applications because `contents` is `iodata` and not a string.

Fixes #334

### DIFF
--- a/lib/mix/tasks/gettext.extract.ex
+++ b/lib/mix/tasks/gettext.extract.ex
@@ -79,12 +79,9 @@ defmodule Mix.Tasks.Gettext.Extract do
   end
 
   defp run_up_to_date_check(pot_files) do
-    not_extracted_paths =
-      for {path, contents} <- pot_files,
-          not File.exists?(path) or File.read!(path) != contents,
-          do: path
+    not_extracted_paths = for {path, _contents} <- pot_files, do: path
 
-    if not_extracted_paths == [] do
+    if pot_files == [] do
       :ok
     else
       Mix.raise("""
--- a/lib/gettext/extractor.ex
+++ b/lib/gettext/extractor.ex
@@ -208,7 +208,11 @@ defmodule Gettext.Extractor do
     # then we can take advantage of Map.merge/3 to find files that we have to
     # update, delete, or add.
     pot_files = Map.new(pot_files, &{&1, :existing})
-    po_structs = Map.new(po_structs)
+
+    po_structs =
+      Map.new(po_structs, fn {path, struct} ->
+        {path, Merger.prune_references(struct, gettext_config)}
+      end)
 
     # After Map.merge/3, we have something like:
     #   %{path => {:merged, :unchanged | %Messages{}}, path => %Messages{}, path => :existing}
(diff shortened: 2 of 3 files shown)


johantell:
Perhaps it can be tested by first creating a temporary file and executing the whole task (forcing the task to read the file from disk like it would do) and capturing the output?

maennchen:
I was able to track the issue down and was able to simplify the code. The issue occured because:

* `Gettext.Extractor` checks if a file changed. To do this is compares the existing and the new POT files: https://github.com/elixir-gettext/gettext/blob/03f9ba7d51da5b45790f0d60c73e73d50664e6b6/lib/gettext/extractor.ex#L221
* It does not call `Merger.prune_references/2`, which causes it to never be the same when `write_reference_comments` is set to `false`
* The references are pruned in https://github.com/elixir-gettext/gettext/blob/03f9ba7d51da5b45790f0d60c73e73d50664e6b6/lib/gettext/extractor.ex#L270
* The file is then reported as changed and passed to the mix task.
* The mix task compares the files again and is comparing `string == iodata` and wlways fails: https://github.com/elixir-gettext/gettext/blob/03f9ba7d51da5b45790f0d60c73e73d50664e6b6/lib/mix/tasks/gettext.extract.ex#L84
* An error is reported

I fixed this by moving the `prune_references` call earlier and making sure the comparising inside `Gettext.Extractor` compares properly. The mix task no longer needs to compare the files since it was exhaustively checked before.

lib/mix/tasks/gettext.extract.ex
```diff
@@ -79,12 +79,9 @@ defmodule Mix.Tasks.Gettext.Extract do
...
 
   defp run_up_to_date_check(pot_files) do
-    not_extracted_paths =
-      for {path, contents} <- pot_files,
-          not File.exists?(path) or File.read!(path) != contents,
-          do: path
+    not_extracted_paths = for {path, _contents} <- pot_files, do: path
 
-    if not_extracted_paths == [] do
+    if pot_files == [] do
```
whatyouhide, line 84: Can you elaborate on this? We're passing the whole list of POT files that we extracted, and if there are any of them then the check fails?

maennchen: @whatyouhide This is not correct:

* If same, return `:unchanged` - https://github.com/jshmrtn/gettext/blob/b0eeb5f05a4d4798f08381685f7bc812e9eaacf0/lib/gettext/extractor.ex#L226
* If `:unchanged`, remove from list - https://github.com/jshmrtn/gettext/blob/b0eeb5f05a4d4798f08381685f7bc812e9eaacf0/lib/gettext/extractor.ex#L208

Therefore the list will only contain entries that have a change. The same was happening before, except that it was not removing references correctly. => Checked twice.

whatyouhide: Ah ok, got it, makes sense! Nice catch. 🙃

whatyouhide (approved):
💯

maennchen:
rebased

maennchen:
oh, seems to have a flaky test. investigating

maennchen:
No more errors with 527f160e3890bb59e58bb82badf3e8bf98f4c1a6

whatyouhide (approved):

maennchen:
@whatyouhide I can’t merge that one though. I only have permission on expo 😊

PR #336, opened 2023-01-09
Make example consistent

nathanl:


whatyouhide (approved):
Great catch, thanks! 👍

PR #337, opened 2023-01-11
Simplify writing of initial pot file top comments

maennchen:
Simplifies the code paths quite a bit and is cleaner since the comments are handled via the `Expo.Messages` structs as well.

I had to remove one assertion from a merge test since the comment is introduced in a different place. It is still covered properly in various places though like https://github.com/elixir-gettext/gettext/blob/03f9ba7d51da5b45790f0d60c73e73d50664e6b6/test/gettext/extractor_test.exs#L51

### DIFF
--- a/lib/gettext/extractor.ex
+++ b/lib/gettext/extractor.ex
@@ -21,6 +21,21 @@ defmodule Gettext.Extractor do
 
   @extracted_messages_flag "elixir-autogen"
 
+  @new_pot_comment String.split(
+                     """
+                     # This file is a PO Template file.
+                     #
+                     # "msgid"s here are often extracted from source code.
+                     # Add new messages manually only if they're dynamic
+                     # messages that can't be statically extracted.
+                     #
+                     # Run "mix gettext.extract" to bring this file up to
+                     # date. Leave "msgstr"s empty as changing them here has no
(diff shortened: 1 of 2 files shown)


lib/gettext/extractor.ex
```diff
@@ -21,6 +21,21 @@ defmodule Gettext.Extractor do
 
   @extracted_messages_flag "elixir-autogen"
 
+  @new_pot_comment String.split(
+                     """
+                     # This file is a PO Template file.
+                     #
+                     # "msgid"s here are often extracted from source code.
+                     # Add new messages manually only if they're dynamic
```
maennchen, line 29: This reverts 09bb5d93e219ace9987ad44cdb3bfee6b695d270 since it was not the cause of #334 and is not "forced" on existing users.

whatyouhide: Let's leave the previous wording, I don't see the reason for changing it. Thanks. 🙃

whatyouhide (changes requested):

lib/gettext/extractor.ex
whatyouhide: Alright, let's standardize 🙄 😉

whatyouhide (approved):

PR #343, opened 2023-01-15
Supply plurals forms header to Gettext.Plural callbacks

### LINKED ISSUE #318: Plural formas parser for pluralization
## Proposed Solution

* Leave default implementation as is
* Allow plugging in a custom pluralization module (already possible)
* Extend `Gettext.Plural`
  - The plural forms header should be passed to both `plurals` & `nplurals`
* Based on our experiences going forward, this behavior could also be made the default if people agree to do so.

## Problem description

`gettext` currently does not parse the `Plural-Forms` header in `.po` / `.mo` files. Instead, it provides an elixir version of default pluralizations. This has a few shortcomings:

### Missing Locales

There are 183 ISO3166 languages, while `gettext` provides 144.

### Mismatch Plural Forms Header / Elixir default

Translation tools, like POEDIT, Crowdin, Transfex, all support and can output of plural-forms headers. If Elixir doesn’t support those headers then it's possible this is a mismatch between the translator's intent and the result. See

* https://docs.transifex.com/formats/gettext
* https://github.com/vslavik/poedit/blob/master/src/language_impl_plurals.h
* https://store.crowdin.com/gnu-gettext/

### XLIFF

IFF is the premium interchange solution for translators and translation tooling and it also embeds plural-rule headers

https://docs.oasis-open.org/xliff/v1.2/xliff-profile-po-1.2-pr-02-20061016-DIFF.pdf

### Primitive Fallback

Gettext has extremely primitive locale fallback mechanisms. It has no proper support for BCP47 language tags. In such a case, its ability to resolve the correct Gettext locale is typically just text equality matching. A totally valid locale of `en_Latn_US` wouldn’t ever match a gettext locale of `en`. In which case, how would a user specify plural forms if they can’t provide a plural-forms header themselves?

Thanks to @kipcole9 for providing me with detailed arguments in favor of this feature.

## Dependencies / Performance

~~As discussed with @josevalim via Slack, we do not want to introduce new dependencies into the `gettext` library other than `expo`. This could be solved either by using `nimble_parsec.compile` and removing `nimble_parsec` as a dependency or to switch to a `yacc` based approach as well.~~

*EDIT: Implemented in https://github.com/elixir-gettext/expo_plural_forms/pull/29*

## Relates to

* https://github.com/elixir-gettext/gettext/pull/313
* https://github.com/elixir-gettext/gettext/pull/306
* https://github.com/elixir-gettext/expo/pull/64
* https://github.com/elixir-gettext/expo_plural_forms

maennchen:
Closes #318

This makes it possible to react precisely to the header instead of using predefined formats.

### DIFF
--- a/lib/gettext/merger.ex
+++ b/lib/gettext/merger.ex
@@ -5,6 +5,7 @@ defmodule Gettext.Merger do
   alias Expo.Message
   alias Expo.Messages
   alias Gettext.Fuzzy
+  alias Gettext.Plural
 
   @new_po_informative_comment """
   # "msgid"s in this file come from POT (.pot) files.
@@ -53,7 +54,7 @@ defmodule Gettext.Merger do
           {Messages.t(), map()}
   def merge(%Messages{} = old, %Messages{} = new, locale, opts)
       when is_binary(locale) and is_list(opts) do
-    opts = put_plural_forms_opt(opts, old.headers, locale)
+    opts = put_plural_forms_opt(opts, old, locale)
     stats = %{new: 0, exact_matches: 0, fuzzy_matches: 0, removed: 0, marked_as_obsolete: 0}
 
     {messages, stats} = merge_messages(old.messages, new.messages, opts, stats)
@@ -220,12 +221,13 @@ defmodule Gettext.Merger do
   """
   def new_po_file(po_file, pot_file, locale, opts) when is_binary(locale) and is_list(opts) do
     pot = PO.parse_file!(pot_file)
-    opts = put_plural_forms_opt(opts, pot.headers, locale)
+    opts = put_plural_forms_opt(opts, pot, locale)
     plural_forms = Keyword.fetch!(opts, :plural_forms)
+    plural_forms_header = Keyword.fetch!(opts, :plural_forms_header)
 
     po = %Messages{
       top_comments: String.split(@new_po_informative_comment, "\n", trim: true),
-      headers: headers_for_new_po_file(locale, plural_forms),
+      headers: headers_for_new_po_file(locale, plural_forms_header),
       file: po_file,
       messages: Enum.map(pot.messages, &prepare_new_message(&1, plural_forms))
     }
@@ -277,11 +279,11 @@ defmodule Gettext.Merger do
     Enum.reject(unique_refs, &match?([], &1))
   end
 
-  defp headers_for_new_po_file(locale, plural_forms) do
+  defp headers_for_new_po_file(locale, plural_forms_header) do
     [
       "",
       ~s(Language: #{locale}\n),
-      ~s(Plural-Forms: nplurals=#{plural_forms}\n)
+      ~s(Plural-Forms: #{plural_forms_header}\n)
     ]
   end
 
@@ -295,21 +297,27 @@ defmodule Gettext.Merger do
     %{message | comments: Enum.reject(comments, &match?("#" <> _, &1))}
   end
 
-  defp put_plural_forms_opt(opts, headers, locale) do
-    Keyword.put_new_lazy(opts, :plural_forms, fn ->
-      read_plural_forms_from_headers(headers) ||
-        Application.get_env(:gettext, :plural_forms, Gettext.Plural).nplurals(locale)
-    end)
-  end
+  defp put_plural_forms_opt(opts, messages, locale) do
+    plural_mod = Application.get_env(:gettext, :plural_forms, Gettext.Plural)
 
-  defp read_plural_forms_from_headers(headers) do
-    Enum.find_value(headers, fn header ->
-      with "Plural-Forms:" <> rest <- header,
-           "nplurals=" <> rest <- String.trim(rest),
-           {plural_forms, _rest} <- Integer.parse(rest) do
-        plural_forms
+    opts =
+      Keyword.put_new_lazy(opts, :plural_forms, fn ->
+        plural_mod.nplurals(Plural.plural_info(locale, messages, plural_mod))
+      end)
+
+    Keyword.put_new_lazy(opts, :plural_forms_header, fn ->
+      requested_nplurals = Keyword.fetch!(opts, :plural_forms)
+
+      default_nplurals = plural_mod.nplurals(Plural.plural_info(locale, messages, plural_mod))
+
+      # If nplurals is overridden to a non-default value by the user the
+      # implementation will not be able to provide a correct header therefore
+      # the header is just set to `nplurals=#{n}` and it is up to the user to
+      # put a complete plural forms header himself.
+      if requested_nplurals == default_nplurals do
+        Plural.plural_forms_header_impl(locale, messages, plural_mod)
       else
-        _other -> nil
+        "nplurals=#{requested_nplurals}"
       end
     end)
   end
--- a/lib/gettext/plural.ex
+++ b/lib/gettext/plural.ex
@@ -122,18 +122,52 @@ defmodule Gettext.Plural do
 
   """
 
+  alias Expo.Messages
+
+  # Types
+
+  @type locale :: String.t()
+
+  @type pluralization_context :: %{
+          optional(:plural_forms_header) => String.t(),
+          required(:locale) => locale()
+        }
+
+  @type plural_info :: term()
+
   # Behaviour definition.
+
+  @doc """
+  Initialize context for `nplurals/1` / `plurals/2`.
+
+  Perform all preparations (for example parsing the plural forms header) per
+  for the provided locale to offload into compile time.
+
+  If the `init/1` callback is not defined, the `plural_info` will be set to the
+  `locale`.
+  """
+  @callback init(pluralization_context()) :: plural_info()
 
   @doc """
   Returns the number of possible plural forms in the given `locale`.
   """
-  @callback nplurals(locale :: String.t()) :: pos_integer
+  @callback nplurals(plural_info()) :: pos_integer()
 
   @doc """
   Returns the plural form in the given `locale` for the given `count` of
   elements.
   """
-  @callback plural(locale :: String.t(), count :: integer) :: plural_form :: non_neg_integer
+  @callback plural(plural_info(), count :: integer()) :: plural_form :: non_neg_integer()
+
+  @doc """
+  Returns the plural forms header value for the given `locale`.
+
+
+  Fallback if not implemented: `"nplurals={nplurals};"`.
+  """
+  @callback plural_forms_header(locale()) :: String.t() | nil
+
+  @optional_callbacks init: 1, plural_forms_header: 1
 
   defmodule UnknownLocaleError do
     @moduledoc """
@@ -434,9 +468,21 @@ defmodule Gettext.Plural do
     "sk"
   ]
 
+  @doc false
+  def init(%{locale: locale, plural_forms_header: plural_forms_header}) do
+    case read_plural_forms_from_headers(plural_forms_header) do
+      nil -> locale
+      nplurals -> {locale, nplurals}
+    end
+  end
+
+  def init(%{locale: locale}), do: locale
+
   # Number of plural forms.
 
   def nplurals(locale)
+
+  def nplurals({_locale, nplurals}), do: nplurals
 
   # All the groupable forms.
 
@@ -510,6 +556,8 @@ defmodule Gettext.Plural do
   # Plural form of groupable languages.
 
   def plural(locale, count)
+
+  def plural({locale, _nplurals}, count), do: plural(locale, count)
 
   # All the `x_Y` languages that have different pluralization rules than `x`.
 
@@ -679,6 +727,71 @@ defmodule Gettext.Plural do
     case String.split(locale, "_", parts: 2, trim: true) do
       [lang, _territory] -> fun.(lang)
       _other -> raise UnknownLocaleError, locale
+    end
+  end
+
+  @doc false
+  def plural_info(locale, messages_struct, plural_mod) do
+    ensure_loaded!(plural_mod)
+
+    if function_exported?(plural_mod, :init, 1) do
+      pluralization_context = %{locale: locale}
+
+      pluralization_context =
+        case Messages.get_header(messages_struct, "Plural-Forms") do
+          [] ->
+            pluralization_context
+
+          plural_forms ->
+            Map.put(
+              pluralization_context,
+              :plural_forms_header,
+              IO.iodata_to_binary(plural_forms)
+            )
+        end
+
+      plural_mod.init(pluralization_context)
+    else
+      locale
+    end
+  end
+
+  @doc false
+  def plural_forms_header_impl(locale, messages_struct, plural_mod) do
+    ensure_loaded!(plural_mod)
+
+    plural_forms_header =
+      if function_exported?(plural_mod, :plural_forms_header, 1) do
+        plural_mod.plural_forms_header(locale)
+      end
+
+    if plural_forms_header do
+      plural_forms_header
+    else
+      nplurals = plural_mod.nplurals(plural_info(locale, messages_struct, plural_mod))
+
+      "nplurals=#{nplurals}"
+    end
+  end
+
+  defp read_plural_forms_from_headers(header) do
+    with "nplurals=" <> rest <- String.trim(header),
+         {plural_forms, _rest} <- Integer.parse(rest) do
+      plural_forms
+    else
+      _other -> nil
+    end
+  end
+
+  # TODO: remove when we depend on Elixir 1.12+
+  if function_exported?(Code, :ensure_loaded!, 1) do
+    defp ensure_loaded!(mod), do: Code.ensure_loaded!(mod)
+  else
+    defp ensure_loaded!(mod) do
+      case Code.ensure_loaded(mod) do
+        {:module, ^mod} -> :ok
+        _other -> raise "<good error message>"
+      end
     end
   end
 end
--- a/lib/mix/tasks/gettext.merge.ex
+++ b/lib/mix/tasks/gettext.merge.ex
@@ -76,8 +76,8 @@ defmodule Mix.Tasks.Gettext.Merge do
   by checking the value of `nplurals` in the `Plural-Forms` header in the existing `.po` file. If
   a `.po` file doesn't already exist and Gettext is creating a new one or if the `Plural-Forms`
   header is not in the `.po` file, Gettext will use the number of plural forms that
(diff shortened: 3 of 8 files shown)


lib/gettext/merger.ex
```diff
@@ -288,22 +288,43 @@ defmodule Gettext.Merger do
...
   end
+
+  defp pluralization_context(messages_struct) do
+    case Messages.get_header(messages_struct, "Plural-Forms") do
+      [] -> %{}
+      plural_forms -> %{plural_forms_header: IO.iodata_to_binary(plural_forms)}
+    end
+  end
+
+  defp read_plural_forms_from_headers(messages) do
```
maennchen, line 320: Why do we have this code? The current default implementation can't handle the header completely and would create an inconsistency.

I think we should probably completely remove this function and delegate all work to `nplurals/2`.

lib/gettext/plural.ex
```diff
@@ -122,18 +122,47 @@ defmodule Gettext.Plural do
 
   """
 
+  # Types
+
+  @type pluralization_context :: %{optional(:plural_forms_header) => String.t()}
```
maennchen, line 127: I've defined the new parameter in a map. That way we can decide to add further context in the future without a breaking change.

whatyouhide:
> shift work into compile time to possibly improve the performance

If we don't know if this improves performance, I don't think we should do it for now. Let's solve it if it ever becomes a problem?

whatyouhide:
@maennchen also, can you elaborate on the purpose of this PR in more details? I'm not really following 😄

maennchen:
@whatyouhide

> If we don't know if this improves performance, I don't think we should do it for now. Let's solve it if it ever becomes a problem?

That highly depends on the implementation of that behaviour. I know that I can make it faster in ExpoPluralForms.

That description should probably be worded differently.

> @maennchen also, can you elaborate on the purpose of this PR in more details? I'm not really following smile

Sure :)

The purpose of this PR is that the plural form header is passed into `Gettext.Plural` implementations.

Therefore implementations can make decisions based on that header.

This will allow it to implement an ExpoPluralForms implementation of the behaviour.

## Current behaviour

Gettext has a list of known locales for which `plural` is implemented. Since that list identifies rule by the locale name, the index is looked up by `locale` & `count`.

## ExpoPluralForms behaviour

For various languages, there's not a single way to represent plural rules.

For example for chinese there's two different rules depending on the context. (`nplurals=1; plural=0;` & `nplurals=2; plural=(n > 1);`)

Also in my own language (swiss german) there's some debate of how to handle `0` counts and depending on the dialect, either new singular case (0) or many (1) applies.

Those rules & differences can be represented using the plural forms header.

By supplying the plural forms header to the callbacks, the implementation can make decisions as they were intended in gettext.

maennchen:
@whatyouhide Performance comparison (extended from https://github.com/elixir-gettext/expo_plural_forms/pull/33#issuecomment-1383212655) to illustrate what would happen if the header was parsed on every index call:

```elixir
Mix.install([{:benchee, "~> 1.1"}, {:expo_plural_forms, path: "."}])

{:ok, arabic} = ExpoPluralForms.Known.plural_form("ar")

defmodule Arabic do
  def index(n), do: unquote(ExpoPluralForms.compile_index(arabic))
end

Benchee.run(
  %{
    "precompiled" => fn ->
      Arabic.index(17)
    end,
    "without" => fn ->
      ExpoPluralForms.index(arabic, 17)
    end,
    "including parse" => fn ->
      {:ok, arabic} = ExpoPluralForms.parse(" nplurals=6; plural=(n==0 ? 0 : n==1 ? 1 : n==2 ? 2 : n%100>=3 && n%100<=10 ? 3 : n%100>=11 ? 4 : 5);")
      ExpoPluralForms.index(arabic, 17)
    end
  },
  time: 10,
  memory_time: 2
)
```

Result:

```
Operating System: Linux
CPU Information: 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
Number of Available Cores: 8
Available memory: 46.77 GB
Elixir 1.14.2
Erlang 25.2

Benchmark suite executing with the following configuration:
warmup: 2 s
time: 10 s
memory time: 2 s
reduction time: 0 ns
parallel: 1
inputs: none specified
Estimated total run time: 42 s

Benchmarking including parse ...
Benchmarking precompiled ...
Benchmarking without ...

Name                      ips        average  deviation         median         99th %
precompiled           12.75 M       78.44 ns ±11373.62%          72 ns         100 ns
without                4.89 M      204.65 ns ±17541.84%         173 ns         348 ns
including parse        0.35 M     2833.62 ns   ±846.85%        2546 ns        4091 ns

Comparison: 
precompiled           12.75 M
without                4.89 M - 2.61x slower +126.21 ns
including parse        0.35 M - 36.12x slower +2755.17 ns

Memory usage statistics:

Name               Memory usage
precompiled                 0 B
without                     0 B - 1.00x memory usage +0 B
including parse          9928 B - ∞ x memory usage +9928 B

**All measurements for memory usage were the same**
```

josevalim:
> By supplying the plural forms header to the callbacks, the implementation can make decisions as they were intended in gettext.

I think this is approaching the problem based on the current solution but, given the limitations you added, I would say the problem needs to be reworked altogether. Likely we want to compile the rules encoded in the header as part of the backend compilation?

When you say the PluralForms header, is that a header for the .po file or specific to a given translation/msgid?

maennchen:
@josevalim The plural forms header is in the `.po` header and not per message.

The idea with `compile_plural` was to compile the rule once per backend and call it from there.

Based on the discussion over in ExpoPluralForms (performance), I think it would make more sense to go a plug like approach. There could be some kind of `init` callback to do preparation works like header parsing and then a normal function call with the init result to calculate the index.

I'll put the PR back into draft state and implement a cleaner solution.

josevalim:
> For example for chinese there's two different rules depending on the context. (nplurals=1; plural=0; & nplurals=2; plural=(n > 1);)

and how do you determine which one to use? :)

maennchen:
@josevalim  The user of the library can decide which rule to use when he sets the header in the `.po` file.

Many tools like [Poedit](https://poedit.net/) help a user setting up the headers as well.

We have an example of that here in the source: https://github.com/elixir-gettext/gettext/blob/2ebe8066d6309ea2d810beb93e38433fef6f5071/test/fixtures/po_editors/poedit.po#L13

whatyouhide:
Yeah, I think the `Plural-Form` header in each PO file is pretty neat. I think compiling the `Plural-Forms` header **once** is a no-brainer, because we can parse it into an Elixir data structure really easily. I think "precompiled" vs "without" in your benchmarks show that "without" is probably a good compromise between simplicity and speed.

maennchen:
@josevalim / @whatyouhide I pushed a new version that is a lot simpler and does not cause any changes in the behaviour. Only a new callback is added.

maennchen:
Example implementation: https://github.com/elixir-gettext/expo_plural_forms/pull/36

lib/gettext/plural.ex
```diff
@@ -434,10 +467,22 @@ defmodule Gettext.Plural do
     "sk"
   ]
 
+  @doc false
+  def init(%{locale: locale, plural_forms_header: plural_forms_header}) do
+    case read_plural_forms_from_headers(plural_forms_header) do
```
maennchen, line 472: This is here to keep `gettext.merge --plural-forms 7` working.

It is however not consistent internally. I could for example merge with `--plural-forms 1` in a language that has 2+ plural forms.

The using `ngettext`, the correct plural forms would not be defined and therefore always result in a missing translation.

---

As an laternate solution, we could remove the option entirely and make the `Gettext.Plural` implementation only work on known formats. If a user wants to add / change a language, he would need to add his own implementation as decribed here: https://github.com/elixir-gettext/gettext/blob/41e16f2f9a84e0b9d8ac77122a623d4d8e7e9a3a/lib/gettext/plural.ex#L41

Even though I personally like option 2 a lot more, I decided to go with the solution that does not cause breaking changes.

lib/gettext/plural.ex
```diff
@@ -122,18 +122,51 @@ defmodule Gettext.Plural do
...
   """
-  @callback nplurals(locale :: String.t()) :: pos_integer
+  @callback nplurals(opts()) :: pos_integer()
 
   @doc """
   Returns the plural form in the given `locale` for the given `count` of
   elements.
   """
-  @callback plural(locale :: String.t(), count :: integer) :: plural_form :: non_neg_integer
+  @callback plural(opts(), count :: integer()) :: non_neg_integer()
```
whatyouhide, line 160: Nit: let's leave the name, it's helpful in documentation.

```suggestion
  @callback plural(opts(), count :: integer()) :: plural_form :: non_neg_integer()
```

lib/gettext/plural.ex
```diff
@@ -122,18 +122,51 @@ defmodule Gettext.Plural do
...
+
+  @doc """
+  Returns the plural forms header value for a given locale
+
+  Fallback if not implemented: `"nplurals={nplurals};"`
```
whatyouhide, line 165:
```suggestion
  Fallback if not implemented: `"nplurals={nplurals};"`.
```

lib/gettext/plural.ex
```diff
@@ -122,18 +122,51 @@ defmodule Gettext.Plural do
...
+  """
+  @callback plural_forms_header(locale()) :: String.t() | nil
```
whatyouhide, line 167: When does this return `nil`, and what does that mean semantically?

lib/gettext/plural.ex
```diff
@@ -122,18 +122,51 @@ defmodule Gettext.Plural do
...
 
   @doc """
   Returns the plural form in the given `locale` for the given `count` of
   elements.
   """
-  @callback plural(locale :: String.t(), count :: integer) :: plural_form :: non_neg_integer
+  @callback plural(opts(), count :: integer()) :: non_neg_integer()
+
+  @doc """
+  Returns the plural forms header value for a given locale
```
whatyouhide, line 163:
```suggestion
  Returns the plural forms header value for the given `locale`.

```

lib/gettext/plural.ex
```diff
@@ -122,18 +122,51 @@ defmodule Gettext.Plural do
...
+  # Types
+
+  @type locale :: String.t()
+
+  @type pluralization_context :: %{
+          optional(:plural_forms_header) => String.t(),
+          required(:locale) => locale()
+        }
+
+  @type opts :: locale() | term()
```
whatyouhide, line 136: Let's rename this to something else, since `opts` doesn't convey what this is. `plural_info`?

Also, why have `locale() | term()` instead of just `term()`?

lib/gettext/plural.ex
```diff
@@ -681,4 +728,52 @@ defmodule Gettext.Plural do
       _other -> raise UnknownLocaleError, locale
     end
   end
+
+  @doc false
+  def pluralization_opts(locale, messages_struct, plural_mod) do
+    {:module, ^plural_mod} = Code.ensure_loaded(plural_mod)
```
whatyouhide, line 734: Should this give a better error message if the `plural_mod` is not loaded, explaining why it needs to be loaded?

lib/mix/tasks/gettext.merge.ex
```diff
@@ -94,9 +94,9 @@ defmodule Mix.Tasks.Gettext.Merge do
       match. Overrides the global `:fuzzy_threshold` option (see the docs for
       `Gettext` for more information on this option).
 
-    * `--plural-forms` - an integer strictly greater than `0`. If this is passed,
-      new messages in the target PO files will have this number of empty
-      plural forms. See the "Plural forms" section above.
+    * `--plural-forms-header` - `Plural-Forms` header content as string.
```
whatyouhide, line 97: This is a breaking change. We can't remove the `--plural-forms` option?

lib/mix/tasks/gettext.merge.ex
```diff
@@ -289,6 +291,10 @@ defmodule Mix.Tasks.Gettext.Merge do
       Mix.raise("The :fuzzy_threshold option must be a float >= 0.0 and <= 1.0")
     end
 
+    unless opts[:plural_forms] == nil do
+      Mix.raise("The :plural_forms option is deprecated in favor of :plural_forms_header")
```
whatyouhide, line 295: This is not a deprecation, if we're raising here. Let's not even deprecate this for now, let's just support both. Deprecations can always be done later.

whatyouhide (changes requested):
Left a bunch of comments!

lib/gettext/plural.ex
maennchen: It means that it does not know a rule for that locale. The caller of the function will then just set a partial header with only nplurals defined.

lib/gettext/plural.ex
maennchen: @whatyouhide I set `opts` based on `plug` semantics: https://github.com/elixir-plug/plug/blob/15445099f5cdd9a196bd77c515e6fb82f1d4801d/lib/plug.ex#L57

In the case where the the optional callback `init/1` is implemented, it can set it to any value it wants. In the case where it is not defined, it will be set to the `locale`.

This is implemented this way so that the `Gettext.Plural` behaviour does not have a breaking change. Otherwise, I would probably not set it as optional or just set it to `nil` if it is not defined.

lib/gettext/plural.ex
maennchen: @whatyouhide I needed to put it into place since it is loaded automatically if a function is called but not to check with `function_exported?`.

We can yield an error though if it cant be loaded.

Something like this?

```suggestion
    case Code.ensure_loaded(plural_mod) do
      {:module, ^plural_mod} -> :ok
      _other -> raise "The module #{plural_mode} can't be found." # Better message?
    end
```

lib/mix/tasks/gettext.merge.ex
maennchen: @whatyouhide It was only supposed to be a deprecation and not a removal except that I used `Mix.raise` by accident.

The idea was that it is deprecated and just yields a warning.

The reason that I wanted to deprecate it is, that it has an inconsistent behaviour as it is right now. (See https://github.com/elixir-gettext/gettext/pull/343#discussion_r1071094099)

maennchen: Based on the information in the comment above:

* Would you like me to not deprecate it and keep it inconsistent?
* Or would you like me to fix the `Mix.raise` by using an actual warning like I intended?

lib/gettext/plural.ex
maennchen: Ignore my previous comment, that is not factually correct. That's what's supposed to happen though. I'll have a look.

maennchen: Now it is as I described.

lib/gettext/plural.ex
josevalim: Maybe we can deprecate the option and guide users towards the new solution.

lib/gettext/plural.ex
maennchen: @josevalim I initially used that one. Unfortunately, this lib supports 1.11 and `ensure_loaded!` was introduced in 1.12.

lib/gettext/plural.ex
whatyouhide: I definitely agree that the pattern of having `init/1` optional and returning a term that then is passed down is great. I'm just thinking that the **name** `opts` is not good. I don't think it's good in Plug either, FWIW 😄 That's why I was suggesting `plural_info`.

As per the typespec, in Plug there is not `term()`, so I think here you can reduce the typespec to just `term()`.

lib/gettext/plural.ex
whatyouhide: @maennchen define `ensure_loaded!/1` in this module conditionally, something like this

```elixir
# TODO: remove when we depend on Elixir 1.12+
if function_exported?(Code, :ensure_loaded!, 1) do
  defp ensure_loaded!(mod), do: Code.ensure_loaded!(mod)
else
  defp ensure_loaded!(mod) do
    case Code.ensure_loaded(mod) do
      {:module, ^mod} -> :ok
      _other -> raise "<good error message>"
    end
  end
end
```

lib/gettext/plural.ex
```diff
@@ -681,4 +729,57 @@ defmodule Gettext.Plural do
...
+  @doc false
+  def plural_forms_header_impl(locale, messages_struct, plural_mod) do
+    {:module, ^plural_mod} = Code.ensure_loaded(plural_mod)
+
+    plural_forms_header =
+      if function_exported?(plural_mod, :plural_forms_header, 1) do
+        plural_mod.plural_forms_header(locale)
+      end
+
+    if plural_forms_header != nil do
```
whatyouhide, line 768: Nitpick

```suggestion
    if plural_forms_header do
```

lib/mix/tasks/gettext.merge.ex
whatyouhide: For now, let's not deprecate and keep it inconsistent. We can do the **soft** deprecation (warning) in a later PR, so that we can keep this PR more focused.

lib/gettext/plural.ex
maennchen: @josevalim That is what I intended to do here: https://github.com/elixir-gettext/gettext/pull/343#discussion_r1081382215

Based on the discussion with @whatyouhide I will probably ignore the whole topic for the PR.

maennchen:
@whatyouhide I've adressed all the comments except https://github.com/elixir-gettext/gettext/pull/343#discussion_r1081221654

Are we sure that we handle the deprecation question correctly? I have removed it but find it a bit strange to keep code (and not even discourage using it) when the behaviour it causes is potentially flawed.

whatyouhide (approved):
This is great for now, we can deprecate the `--plural-forms` option later. 💯

maennchen:
@whatyouhide / @josevalim 
Quick note: I will have some surgery on my wrist next week. I will probably not have time to finish the Gettext.Plural implementation for expo until then.
I however plan to pick it back up 2/3 weeks later when I can hopefully get rid of the cast and properly type again.

whatyouhide:
@maennchen no worries! I have some free time, so I can give it a shot if you want. I'm working on polishing docs as well. Thanks for all the work!! 💟

maennchen:
@whatyouhide If you’d like to, I don’t mind if you pick it up from here. But otherwise I’m also fine doing myself in a few weeks. However you prefer 😊 I might be able to do some code reviews, so just let me know if I can help.

PR #346, opened 2023-01-20
fix: check-up-to-date check error

janwillemvd:
We started getting the error below:

Screenshot 2023-01-20 at 16 55 40

It might have been introduced after our recent upgrade to version `0.21.0`.


maennchen:
@janwillemvd Thanks!

I believe that this has already been solved but not yet released via #335.

If that is not true: Do you have a test-case to reproduce the error you get when using the latest `main`?

janwillemvd:
Thanks, @maennchen. Didn't see #335. Just checked and with the latest master there is indeed no problem anymore :)

PR #347, opened 2023-01-21
Use Expo for handling built-in plural forms

whatyouhide:


mix.exs
```diff
@@ -49,7 +49,8 @@ defmodule Gettext.Mixfile do
 
   defp deps do
     [
-      {:expo, "~> 0.1.0 or ~> 0.2.0"},
+      # {:expo, "~> 0.1.0 or ~> 0.2.0"},
+      {:expo, github: "elixir-gettext/expo"},
```
whatyouhide, line 53: We'll update this once we release a new Expo version.

lib/gettext/plural.ex
```diff
@@ -256,282 +256,25 @@ defmodule Gettext.Plural do
...
-    with "nplurals=" <> rest <- String.trim(plural_forms_header),
-         {plural_forms, _rest} <- Integer.parse(rest) do
-      {locale, plural_forms}
-    else
-      _other -> locale
+    case Expo.PluralForms.parse(plural_forms_header) do
+      {:ok, plural_forms} ->
+        {locale, plural_forms}
+
+      {:error, _reason} ->
```
maennchen, line 269: @whatyouhide Is a plural forms header with only the nplurals part technically valid? If yes we should maybe parse it to a plural forms struct with a nil plural field.

maennchen (approved):

lib/gettext/plural.ex
maennchen: I'm trying to find an answer to this in the official gettext source.

lib/gettext/plural.ex
```diff
@@ -256,282 +256,25 @@ defmodule Gettext.Plural do
...
+        # Fall back to parsing headers such as "nplurals=3", without the "plural=..." part.
+        # TODO: remove this at some point.
+        with "nplurals=" <> rest <- String.trim(plural_forms_header),
+             {plural_forms, _rest} <- Integer.parse(rest) do
+          {locale, plural_forms}
```
maennchen, line 274: Would it make sense to put something along the lines of this (untested pseudo-code):

```suggestion
          case Expo.PluralForms.plural_form(locale) do
            {:ok, plural_form} -> IO.warn("""
            Deprecated `Plural-Forms` header found:
            Current: #{inspect(plural_forms_header)}
            Recommendation: #{inspect(Expo.PluralForms.to_string(plural_form))}
            """)
            :error -> :ok
          end
          {locale, plural_forms}
```

lib/gettext/plural.ex
maennchen: I can't find anything. Let's keep it as is.

whatyouhide: @maennchen yeah I don't think it's valid and I'd rather err on the side of not being valid, so I hacked this together to support it only in Gettext for now (and not in Expo). We can remove this support later on 🙃

lib/gettext/plural.ex
whatyouhide: Yes! I'll do this in another PR, where we can deprecate `--plural-forms` as well and so on.

PR #348, opened 2023-01-21
Set correct error message for ensure_loaded! in Plural

maennchen:
Oops, that should not have been merged in #343 

Copied from https://github.com/elixir-lang/elixir/blob/df056c2cda6c046839d42c830888c79c18287bfd/lib/elixir/lib/code.ex#L1648


whatyouhide (approved):

whatyouhide:
Ah yep of course, good catch 👍

PR #349, opened 2023-01-21
Deprecate `nplurals=x` and `:plural_forms`

whatyouhide:


lib/gettext/merger.ex
```diff
@@ -299,21 +329,17 @@ defmodule Gettext.Merger do
 
   defp put_plural_forms_opt(opts, messages, locale) do
     plural_mod = Application.get_env(:gettext, :plural_forms, Gettext.Plural)
+    default_nplurals = plural_mod.nplurals(Plural.plural_info(locale, messages, plural_mod))
 
-    opts =
-      Keyword.put_new_lazy(opts, :plural_forms, fn ->
-        plural_mod.nplurals(Plural.plural_info(locale, messages, plural_mod))
-      end)
+    opts = Keyword.put_new(opts, :plural_forms, default_nplurals)
```
maennchen, line 334: Should we add a `TODO` to simplify the code here in `v0.24.0`?

(`put_new` => `put` and remove `if requested_nplurals == default_nplurals do` condition)

lib/gettext/plural.ex
```diff
@@ -268,9 +268,18 @@ defmodule Gettext.Plural do
...
              {plural_forms, _rest} <- Integer.parse(rest) do
+          IO.warn("""
+          Plural-Forms headers in the form "nplurals=<int>" (without the "plural=<rule>" part \
+          following) are invalid and support for them will be removed in future Gettext \
+          versions. Make sure to use a complete Plural-Forms header, which also specifies \
+          the pluralization rules, or remove the Plural-Forms header completely. If you \
+          do the latter, Gettext will use its built-in pluralization rules for the languages \
+          it knows about (see Gettext.Plural).\
+          """)
+
```
maennchen, line 282: We could directly output the recommended header here to make the transition easier on people.

maennchen: (Removal is actually not a god idea since other tools also depend on it...)

whatyouhide: From where we would output it?

maennchen: @whatyouhide I would probably do that right here. That should pop up at compile time & when merging.

whatyouhide: Added!

maennchen: Love it!

maennchen (approved):

PR #353, opened 2023-02-16
fixes compiler plural_forms_fun name interpolation

### LINKED ISSUE #351: Gettext.PluralFormError for plural form "1" in "ja" locale
I'm a bit confused about plural forms. According to the [GNU Gettext documentation](https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html), the appropriate header for languages that don't distinguish between singular and plural is `Plural-Forms: nplurals=1; plural=0;`. So I added this to my `priv/gettext/ja/LC_MESSAGES/default.po` file:

```
msgid ""
msgstr ""
"Language: ja\n"
"Plural-Forms: nplurals=1; plural=0;\n"
```

Now when I run `mix gettext.extract --merge`, it adds this to the `.po` file for strings passed to `ngettext/3`:

```
msgid "There is a thing."
msgid_plural "There are %{count} things."
msgstr[0] ""
```

After I add a translation:

```
msgid "There is a thing."
msgid_plural "There are %{count} things."
msgstr[0] "ことがある"
```

I'm getting this error:

```
** (Gettext.PluralFormError) plural form 1 is required for locale "ja" but is missing for message compiled from priv/gettext/ja/LC_MESSAGES/default.po:1407
    (my_app 0.1.0) lib/my_app_web/gettext.ex:1: MyAppWeb.Gettext.ja_default_lngettext/5
    (gettext 0.22.0) lib/gettext.ex:931: Gettext.dpngettext/7
    (my_app 0.1.0) lib/my_app_web/live/thing_live/show.ex:219: anonymous fn/2 in MyWeb.ThingLive.Show.notification/1
```

Am I doing something wrong?

gettext 0.22.0

maennchen:
fixes #351

### DIFF
--- a/lib/gettext/compiler.ex
+++ b/lib/gettext/compiler.ex
@@ -504,7 +504,7 @@ defmodule Gettext.Compiler do
     %{locale: locale, domain: domain, path: path} = po_file
     %Messages{messages: messages, file: file} = messages_struct = PO.parse_file!(path)
 
-    plural_forms_fun = :"{locale}_#{domain}_plural"
+    plural_forms_fun = :"#{locale}_#{domain}_plural"
 
     plural_forms = compile_plural_forms(locale, messages_struct, plural_mod, plural_forms_fun)
     nplurals = nplurals(locale, messages_struct, plural_mod)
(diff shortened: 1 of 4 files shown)


whatyouhide:
Great catch! 💟

Issue #358, opened 2023-03-14
Module is not loaded because :nofile


TeoDiaz:
When I compile the project I'm having this error in the last versions, `0.22.0` and `0.22.1`:

```
** (ArgumentError) could not load module MyApp.Gettext.Plural due to reason :nofile
    (elixir 1.14.3) lib/code.ex:1581: Code.ensure_loaded!/1
    (gettext 0.22.0) lib/gettext/plural.ex:362: Gettext.Plural.plural_info/3
    (gettext 0.22.0) lib/gettext/compiler.ex:571: Gettext.Compiler.compile_plural_forms/4
    (gettext 0.22.0) lib/gettext/compiler.ex:509: Gettext.Compiler.compile_po_file/5
    (gettext 0.22.0) lib/gettext/compiler.ex:450: Gettext.Compiler.compile_unified_po_file/4
    (elixir 1.14.3) lib/enum.ex:1658: Enum."-map/2-lists^map/1-0-"/2
    (gettext 0.22.0) expanding macro: Gettext.Compiler.__before_compile__/1
    lib/myapp/gettext.ex:1: MyApp.Gettext (module)
``` 

[This line](https://github.com/elixir-gettext/gettext/blob/main/lib/gettext/plural.ex#L414)

We have configured the Plurals in the MyApp.Gettext module
```
use Gettext, otp_app: :myapp, plural_forms: MyApp.Gettext.Plural
```

I tried to move the configuration to the `config/config.exs` with no luck. 

Maybe I'm missing something, but could this be a bug in the last Plural updates? 

Thanks.

josevalim:
We should use `Code.ensure_compiled!` instead, as it may not have been defined.

whatyouhide:
What @josevalim said, yep.

@TeoDiaz good catch. Do you want to send a PR to fix this? 🙃 💟

TeoDiaz:
I'm already on it! Thanks!

TeoDiaz:
I just changed the exact point to fix this issue, but maybe we could improve it changing the code to the `ensure_compiled!/1` function. 

```
if function_exported?(Code, :ensure_compiled!, 1) do
    defp ensure_compiled!(mod), do: Code.ensure_compiled!(mod)
  else
    defp ensure_compiled!(mod) do
      case Code.ensure_compiled(mod) do
        {:module, ^mod} ->
          mod

        {:error, reason} ->
          raise ArgumentError,
[...]

PR #359, opened 2023-03-14
Change from ensure_loaded/1 to ensure_compiled/1

TeoDiaz:
Fixes https://github.com/elixir-gettext/gettext/issues/358

We can't use ensure_compiled!/1 for versions under 1.12+

### DIFF
--- a/lib/gettext/plural.ex
+++ b/lib/gettext/plural.ex
@@ -368,7 +368,7 @@ defmodule Gettext.Plural do
 
   @doc false
   def plural_info(locale, messages_struct, plural_mod) do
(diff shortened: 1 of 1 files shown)


maennchen (approved):

josevalim:
@TeoDiaz we should rename the function to `ensure_compiled!` and we should also change it to `ensure_compiled!` on the `if` branch.

josevalim:
I pushed one commit but I left the renaming to you. :)

TeoDiaz:
> I pushed one commit but I left the renaming to you. :)

Done! Thanks 😉

PR #368, opened 2023-06-27
Don't rely on the the last-update timestamp of the manifest for extract mix task

### LINKED ISSUE #367: Running `mix gettext.extract` doesn't extract new messages with Elixir 1.15
I'm seeing an issue on new Phoenix projects that run on elixir 1.15. When the `mix gettext.extract` task is run, it doesn't pick up new messages anymore (nor does the same mix task with the `--check-up-to-date` option signal missing messages).

Removing the `_build` folder triggers the task to pick new messages up again, so it appears to be a cache/recompilation problem.

Reproducing is straightforward:
- install and use elixir 1.15 (with asdf: `asdf install elixir 1.15.0-otp-25 && asdf global elixir 1.15.0-otp-25`)
- I've tested with latest Phoenix installer version (v1.7.6), so ensure this version is installed with `mix archive.install hex phx_new 1.7.6`
- Create a new default Phoenix project
- Run gettext extract mix task: `mix gettext.extract --merge` (this initialises all po/t files)
- Change or add any call to `gettext/1` (there are some calls in the `core_components.ex` file).
- Run gettext extract mix task: `mix gettext.extract --merge`

The last step should result in a new message being extracted to the po/t files, but that's not the case anymore. Running the same procedure with elixir 1.14 _does_ extract a new message.

linusdm:
Don't rely on the the last-update timestamp of the manifest to trigger a forced compilation with the Elixir compiler.

The Mix task needs to be re-run, to ensure that previous compilations don't turn this in a noop. This assumes the `compiler.elixir` is sufficient for extracting messages.

Closes #367.

Needs thorough review though, I'm not sure of all the consequences.

### DIFF
--- a/lib/mix/tasks/gettext.extract.ex
+++ b/lib/mix/tasks/gettext.extract.ex
@@ -113,7 +113,7 @@ defmodule Mix.Tasks.Gettext.Extract do
 
   defp force_compile do
     Mix.Tasks.Compile.Elixir.clean()
-    Enum.each(Mix.Tasks.Compile.Elixir.manifests(), &File.rm/1) 
+    Enum.each(Mix.Tasks.Compile.Elixir.manifests(), &File.rm/1)
 
     # If "compile" was never called, the reenabling is a no-op and
     # "compile.elixir" is a no-op as well (because it wasn't reenabled after
--- a/lib/mix/tasks/gettext.extract.ex
+++ b/lib/mix/tasks/gettext.extract.ex
@@ -112,7 +112,7 @@ defmodule Mix.Tasks.Gettext.Extract do
   end
 
   defp force_compile do
-    Enum.map(Mix.Tasks.Compile.Elixir.manifests(), &make_old_if_exists/1)
+    Enum.map(Mix.Tasks.Compile.Elixir.manifests(), &File.rm/1)
 
     # If "compile" was never called, the reenabling is a no-op and
     # "compile.elixir" is a no-op as well (because it wasn't reenabled after
@@ -122,10 +122,6 @@ defmodule Mix.Tasks.Gettext.Extract do
     Mix.Task.reenable("compile.elixir")
     Mix.Task.run("compile")
     Mix.Task.run("compile.elixir", ["--force"])
-  end
-
-  defp make_old_if_exists(path) do
-    :file.change_time(path, {{2000, 1, 1}, {0, 0, 0}})
   end
 
   defp run_merge(pot_files, argv) do
--- a/lib/mix/tasks/gettext.extract.ex
+++ b/lib/mix/tasks/gettext.extract.ex
@@ -112,7 +112,7 @@ defmodule Mix.Tasks.Gettext.Extract do
   end
 
   defp force_compile do
-    Enum.map(Mix.Tasks.Compile.Elixir.manifests(), &File.rm/1)
+    Mix.Tasks.Compile.Elixir.clean()
 
     # If "compile" was never called, the reenabling is a no-op and
     # "compile.elixir" is a no-op as well (because it wasn't reenabled after
--- a/lib/mix/tasks/gettext.extract.ex
+++ b/lib/mix/tasks/gettext.extract.ex
@@ -113,6 +113,7 @@ defmodule Mix.Tasks.Gettext.Extract do
 
   defp force_compile do
     Mix.Tasks.Compile.Elixir.clean()
+    Enum.each(Mix.Tasks.Compile.Elixir.manifests(), &File.rm/1) 
 
     # If "compile" was never called, the reenabling is a no-op and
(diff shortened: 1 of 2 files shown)


whatyouhide:
Hey @linusdm, great find and thanks for the great reproducing steps in #367. I think the solution that @josevalim mentioned in https://elixirforum.com/t/elixir-v1-15-0-released/56584/71 is closer to what we have today. Can you give that a try insterad?

josevalim:
Ah, my suggestion is not going to work because it is touching the Elixir manifests and those manifests are private. However, since we want to force compile, maybe we could just remove the manifests?

whatyouhide:
Removing the manifests also works. I can't quite remember *why* we did this thing with the manifests, when the solution in this PR (`mix compile.elixir --force`) seems to work on a brand new Phoenix project 🤔

josevalim:
The issue is that we need to make these scenarios work:

* `rm -rf _build; mix gettext.extract --merge` - in particular, we need to run all compilers before extracting, so we can't only run the Elixir one

* `mix do compile + gettext.extract --merge` - the project has been compiled and you want to compile the Elixir parts, not the whole thing

whatyouhide:
Coolsies, @linusdm we have a path forward then! 🙃

linusdm:
Okay, those are some clear boundaries and conditions you have put forward here. I'll take a stab at it, I think I can manage.

Just for clarity and my own understanding: I see how just running `mix compile.elixir --force` fails in the case there was no previous compilation. I don't really understand why, but I checked, and indeed running only the elixir compiler in  isolation from scratch doesn't work. That's due to a lack of my understanding of how compilation works, I guess.
Then I thought rerunning `mix compile --force` would work. But indeed, when combined as mentioned in your second scenario, with `mix do compile + gettext.extract --merge`, that somehow doesn't work. That is totally surprising to me, because I'm explicitly _rerunning_ the task, which in my mind should be totally oblivious to the fact that you explicitly ran a compilation before.
So I can confirm that the current PR is not dealing with both scenarios described above, not even when rerunning a full compilation, instead of just the elixir one.

I'll take your suggestions and remove the manifest file completely, and see how that turns out. I'll try to capture both scenarios in tests, so we don't have to have this discussion again :-) Afaik there are no tests capturing this knowledge yet, as the tests all run fine with this faulty PR.

Btw, there is some more history about this here: https://github.com/elixir-gettext/gettext/issues/104 But I guess you already had your memories refreshed :-)

Thanks for the feedback!

linusdm:
I have tested manually, and it seems to work!
I will add tests later when I have more time (in a separate PR, if you want to merge this already).

linusdm:
Oh, this seems to generate a lot of warnings of the type:

> warning: redefining module FonarWeb.Cldr.Territory (current version loaded from _build/dev/lib/my_app/ebin/Elixir.MyAppWeb.Cldr.Territory.beam)
> lib/my_app_web/cldr.ex:1: MyAppWeb.Cldr.Territory (module)

I didn't see them in my small repro project, but now on our real project, with many dependencies, it seems to generate the warnings for each referenced dependency module.
This seems a bit fishy, and probably has to do with the way the compilations are run, in combination with removing the manifest file.

lib/mix/tasks/gettext.extract.ex
```diff
@@ -112,7 +112,7 @@ defmodule Mix.Tasks.Gettext.Extract do
   end
 
   defp force_compile do
-    Enum.map(Mix.Tasks.Compile.Elixir.manifests(), &make_old_if_exists/1)
+    Enum.map(Mix.Tasks.Compile.Elixir.manifests(), &File.rm/1)
```
josevalim, line 115:
```suggestion
    Mix.Tasks.Compile.Elixir.clean()
```

josevalim:
Gah, you are right. That's the reason why we changed the manifest. I pushed another commit. Can you please give it a try instead?

lib/mix/tasks/gettext.extract.ex
```diff
@@ -112,7 +112,7 @@ defmodule Mix.Tasks.Gettext.Extract do
   end
 
   defp force_compile do
-    Enum.map(Mix.Tasks.Compile.Elixir.manifests(), &make_old_if_exists/1)
+    Mix.Tasks.Compile.Elixir.clean()
```
whatyouhide, line 115: Oh this is great. I wonder why we didn't do that earlier... 😄

josevalim, line 115:
```suggestion
    Mix.Tasks.Compile.Elixir.clean()
    Enum.each(Mix.Tasks.Compile.Elixir.manifests(), &File.rm/1) 
```

josevalim:
Please try again. `clean` was not deleting the manifest.

linusdm:
Yup, that seems to work well in all possible scenarios!

PR #370, opened 2023-07-31
Keep custom flags with the `custom_flags_to_keep` option

### LINKED ISSUE #369: Retain custom flags during merge
**Background:**
In our project, we've implemented an auto-translate feature that helps team members translate to languages they don't know (and we'll help adjust them through the review process or send them over to someone who helps us with the translation).

We've set up a mix task that back-fills all empty translations in a specific language with translations from an external language model and adds the `auto-translated` tag to the message.

We've noticed recently that all those tags get removed when doing a `mix gettext.extract --merge`, which causes some disruption for us as we use those tags to know which messages we should send to our external human translators for verification.

**Problem**
From my understanding (by looking through the `Merger` module), `fuzzy` is the only flag allowed to be retained over multiple merges (https://github.com/elixir-gettext/gettext/blob/main/lib/gettext/merger.ex#L230-L236), which is why we're experiencing our problem

**Suggested solution**
I understand that there is most likely a very sound reason to resetting the tags on each extraction so my suggestion is that we make the retained titles configurable through our mix-file:

```elixir
def project do
  [
    #...
    gettext: [
      retain_flags: ["fuzzy", "auto-translated"]
    ]
    #...
  ]
end
```

Where we either use `["fuzzy"]` as a default or merge the "internal" flags with the configured value

---

If this suggestion sounds good to you, I'd be happy to implement it and shoot a PR your way :)

johantell:
Background:
While merging translations with `mix gettext.merge` flags get cleared out and replaced with the exception of `fuzzy`, which is kept.

This change will allow the `custom_flags_to_keep` flag to add a list of flags to be kept in addition to `fuzzy` which helps prevent information loss if external tools adds their own flags to messages.

Fixes #369


johantell:
It got a little bit awkward to pass on the `gettext_config` as deep down as I had to, but I'll let you decide if it's OK or not :)

whatyouhide:
@johantell I’m seeing that we don't really have any Gettext-level config related to merging, right? Because if so, we _could_ potentially start with making this a CLI flag for the `gettext.merge` task. However, I recognize that it would be easier to configure this on the project level so that contributors to a project don't have to remember to pass the flag. On the third hand, Mix aliases... Thoughts? 😄

lib/gettext.ex
```diff
@@ -586,6 +586,9 @@ defmodule Gettext do
       have a matching reference. You can use this pattern to prevent Gettext from
       removing messages that you have extracted using another tool.
 
+    * `:custom_flags_to_keep` - a list of custom flags that will be kept for
+      existing messages during a merge.
```
whatyouhide, line 590:
```suggestion
    * `:custom_flags_to_keep` - a list of custom flags that will be kept for
      existing messages during a merge. Gettext always keeps the `elixir-format`
      and `fuzzy` flags.
      
```

test/gettext/merger_test.exs
```diff
@@ -504,6 +524,36 @@ defmodule Gettext.MergerTest do
       assert stderr =~ "warning"
       assert stderr =~ "The --plural-forms and :plural_forms options are deprecated"
     end
+
+    test "custom flags defined by `custom_flag_to_keep` config are kept" do
```
whatyouhide, line 528:
```suggestion
    test "custom flags defined by :custom_flag_to_keep config are kept" do
```

whatyouhide (approved):

whatyouhide:
Thank you @johantell! ❤️

lib/gettext/merger.ex
```diff
@@ -202,37 +205,49 @@ defmodule Gettext.Merger do
...
     }
   end
 
-  defp merge_flags(old, new) do
-    if Message.has_flag?(old, "fuzzy") do
-      Message.append_flag(new, "fuzzy").flags
-    else
-      new.flags
-    end
+  @default_flags_to_keep ["elixir-format", "fuzzy"]
```
maennchen, line 237: We should only have a special rule for the fuzzy flag. The format flag is depending on the enabled interpolation module and might differ from elixir-format.

maennchen: Too late for merge, please still consider it @whatyouhide 😊

whatyouhide: @maennchen ah that's a great call out. PR? 😬

maennchen: @whatyouhide I recently had surgery and am on a lot of pain meds. I think it would be better if you could do this without me…

whatyouhide: @maennchen ah, sure! Good luck with recovery then 🤗

whatyouhide: Done! ✔️

whatyouhide:
@johantell I don't have cycles to spend on this right now, so for now no release. I can try to do it by end of week, but we'll see how things look like. Sorry!

johantell:
@whatyouhide Alright, no problem. I'll lock on the latest commit until you've found time. Thanks for taking the time to answer, and let me know if there is anything I can do to help out :)

whatyouhide:
@johantell ok I snuck this in, 0.23.0 is out! 👍

johantell:
@whatyouhide You're my hero! ❤️  💛 💚 💙

PR #382, opened 2023-12-04
Release `v0.24.0`

maennchen:
I though it makes sense to highlight the way to fix a `Expo.PO.DuplicateMessagesError` directly in the changelog.

I would classify this is a patch release since we handled the messages differently from GNU gettext. Do you agree @whatyouhide or should this be `v0.24.0`?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Changelog
 
+## v0.24.0
+
+  * Handle singular and plural messages with the same `msgid` as the same
+    message.
+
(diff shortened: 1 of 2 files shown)


whatyouhide:
@maennchen changelog looks great, but let's do 0.24.0 since this can potentially be a bit disruptive.

whatyouhide (approved):

whatyouhide:
I set this up so that if you publish a tag then GH Actions publishes to Hex right away (https://github.com/elixir-gettext/gettext/actions/runs/7088176540). However, you had forgotten to run `mix deps.get` so I had to update `mix.lock`.

PR #383, opened 2023-12-07
Prevent Unlocked / Outdated Mix Dependencies in CI

maennchen:
Should prevent https://github.com/elixir-gettext/gettext/pull/382#issuecomment-1838814568


whatyouhide (approved):

whatyouhide:
This is awesome, TIL. I didn't know about `--check-locked` 🙃 I'll go do that to all my OSS libraries ahahah.

Issue #330, opened 2022-12-20
Remove compile time depedency of Gettext backend module


maennchen:
Continued discussion of https://github.com/elixir-gettext/gettext/issues/317#issuecomment-1356450733

Removing the compile time dependency to the gettext backend module could speed up compile performance of larger apps considerably.

This has an even bigger impact when used together with `ex_cldr` which also recompiles if it is linked to the gettext backend. (Setting `gettext` in configuration - https://hexdocs.pm/ex_cldr/1.6.4/readme.html#configuration)

maennchen:
@josevalim 

> [...] does not change the code generated at compile-time

Is that technically correct? We're using `Gettext.Interpolation.compile_interpolate/3` to interpolate bindings known at the module's compile time. Therefore the compilation of the module using a translation can depend on the specific compilation string if I understand it correctly.

However: It could make sense to disable `compile_interpolate/3` and only use `runtime_interpolate/2` instead in the `dev` / `test` env. That way, we would get no compile dependency in the environments with frequent changes and the current fast precompiled translations in `prod`. (I would not decide based on the `Mix.env()` but rather add a setting to the `gettext` configuration.

maennchen:
Actually ignore my last comment. We’re not considering the usage of the translation / given bindings. We’re just optimizing the use case where all bindings were correctly supplied.

maennchen:
One solution to the problem I can think of:

* Instead of using macros to extract messages, a compile tracer could be used instead. (`imported_function` & `remote_function`)
* That way the gettext module no longer has to be imported and we can get rid of the compile time depdendency.

One problem with that is approach is that the event do not expose the raw arguments (only the arity). It seems like that could be adapted if we wanted to. I did a small POC some time ago that uses a compile tracer for extraction. (Unfortunately I can't locate it anymore.)  With a few small changes to the compile tracing I was able to get it running.

The reason why I'm thinking about that approach is how elixir handles recompilation:

Given the following code, touching `priv/test` will compile the parent as well:

```elixir
defmodule CompileTest do
  IO.inspect "Compiling outer"
  defmodule CompileTest.Inner do
    IO.inspect "Compiling inner"
    @external_resource Application.app_dir(:compile_test, "priv/test")
  end  
end
```

This also applies if the modules are defined next to each other in the same file.

josevalim:
[...]

PR #390, opened 2024-08-17
Introduce "use Gettext.Backend"

whatyouhide:
This starts to address #330 by allowing `use Gettext.Backend` in place of `use Gettext`. Small steps, and this should be easy to review 🙃 

I left all the test modules to `use Gettext` because we'll change that later on anyway.

### DIFF
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -78,11 +78,11 @@ defmodule Gettext do
 
   ## Gettext API
 
-  To use `Gettext`, a module that calls `use Gettext` (referred to below as a
+  To use `Gettext`, a module that calls `use Gettext.Backend` (referred to below as a
   "backend") has to be defined:
 
       defmodule MyApp.Gettext do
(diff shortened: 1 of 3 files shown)


lib/gettext.ex
```diff
@@ -627,55 +627,15 @@ defmodule Gettext do
...
-          ) do
-        Gettext.Compiler.warn_if_domain_contains_slashes(domain)
-        string = if n == 1, do: msgid, else: msgid_plural
-        bindings = Map.put(bindings, :count, n)
-
-        with {:ok, interpolated} <- @interpolation.runtime_interpolate(string, bindings),
-             do: {:default, interpolated}
+      if Keyword.has_key?(opts, :backend) do
+        raise "not implemented yet"
+      else
```
josevalim, line 635:
```suggestion
      else
        # TODO: Deprecate this branch
```

josevalim (approved):

PR #391, opened 2024-08-17
Avoid compile-time dependency to Gettext backend

whatyouhide:
First pass, let's see if this is a direction we like.

### DIFF
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -634,14 +634,14 @@ defmodule Gettext do
         opts
       end
 
-    case Keyword.fetch(opts, :backend) do
+    case Keyword.keyword?(opts) && Keyword.fetch(opts, :backend) do
       {:ok, backend} ->
         quote do
           @__gettext_backend__ unquote(backend)
           import Gettext.Macros
         end
 
-      :error ->
+      _other ->
         quote do
           # TODO: Deprecate this branch
           use Gettext.Backend, unquote(opts)
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -627,17 +627,34 @@ defmodule Gettext do
 
   @doc false
   defmacro __using__(opts) do
-    quote do
-      opts = unquote(opts)
-
-      if Keyword.has_key?(opts, :backend) do
-        raise "not implemented yet"
+    opts =
+      if Macro.quoted_literal?(opts) do
+        Macro.prewalk(opts, &expand_alias(&1, __CALLER__))
       else
-        # TODO: Deprecate this branch
-        require Gettext.Backend
-        Gettext.Backend.__using__(opts)
+        opts
       end
+
+    case Keyword.fetch(opts, :backend) do
+      {:ok, backend} ->
+        quote do
+          @__gettext_backend__ unquote(backend)
+          import Gettext.Macros
+        end
+
+      :error ->
+        quote do
+          # TODO: Deprecate this branch
+          use Gettext.Backend, unquote(opts)
+        end
     end
+  end
+
+  defp expand_alias({:__aliases__, _, _} = als, env) do
+    Macro.expand(als, %{env | function: {:__gettext__, 1}})
+  end
+
+  defp expand_alias(other, _env) do
+    other
   end
 
   @doc """
--- a/lib/gettext/macros.ex
+++ b/lib/gettext/macros.ex
@@ -1,6 +1,37 @@
 defmodule Gettext.Macros do
-  @moduledoc false
+  @moduledoc """
+  Macros used by Gettext to provide the gettext family of functions.
 
+  *Available since v0.26.0.*
+
+  Macros enable users to use gettext and get **automatic extraction** of translations.
+  See `Gettext` for more information.
+
+  This module is *imported* every time you call:
+
+      use Gettext, backend: MyApp.Gettext
+
+  > #### Warning {: .error}
+  >
+  > You are not meant to use this module in any way other than with `use Gettext` as shown above,
+  > as its macros depend on internals that get set up when you call `use Gettext`.
+  """
+
+  @moduledoc since: "0.26.0"
+
+  @doc """
+  Marks the given message for extraction and returns it unchanged.
+
+  This macro can be used to mark a message for extraction when `mix
+  gettext.extract` is run. The return value is the given string, so that this
+  macro can be used seamlessly in place of the string to extract.
+
+  ## Examples
+
+      dpgettext_noop("errors", "Home page", "Error found!")
+      #=> "Error found!"
+
+  """
   defmacro dpgettext_noop(domain, msgctxt, msgid) do
     domain = expand_domain(domain, __CALLER__)
     msgid = Gettext.Compiler.expand_to_binary(msgid, "msgid", __MODULE__, __CALLER__)
@@ -20,12 +51,38 @@ defmodule Gettext.Macros do
     msgid
   end
 
+  @doc """
+  Marks the given message for extraction and returns it unchanged.
+
+  This macro can be used to mark a message for extraction when `mix
+  gettext.extract` is run. The return value is the given string, so that this
+  macro can be used seamlessly in place of the string to extract.
+
+  ## Examples
+
+      dgettext_noop("errors", "Error found!")
+      #=> "Error found!"
+
+  """
   defmacro dgettext_noop(domain, msgid) do
     quote do
       unquote(__MODULE__).dpgettext_noop(unquote(domain), nil, unquote(msgid))
     end
   end
 
+  @doc """
+  Marks the given message for extraction and returns it unchanged.
+
+  This macro can be used to mark a message for extraction when `mix
+  gettext.extract` is run. The return value is the given string, so that this
+  macro can be used seamlessly in place of the string to extract.
+
+  ## Examples
+
+      gettext_noop("Error found!")
+      #=> "Error found!"
+
+  """
   defmacro gettext_noop(msgid) do
     quote do
       unquote(__MODULE__).dpgettext_noop(
@@ -36,6 +93,19 @@ defmodule Gettext.Macros do
     end
   end
 
+  @doc """
+  Marks the given message for extraction and returns it unchanged.
+
+  This macro can be used to mark a message for extraction when `mix
+  gettext.extract` is run. The return value is the given string, so that this
+  macro can be used seamlessly in place of the string to extract.
+
+  ## Examples
+
+      pgettext_noop("Error found!", "Home page")
+      #=> "Error found!"
+
+  """
   defmacro pgettext_noop(msgid, context) do
     quote do
       unquote(__MODULE__).dpgettext_noop(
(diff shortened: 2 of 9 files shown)


lib/gettext/macros.ex
```diff
@@ -0,0 +1,239 @@
...
+    else
+      quote do
+        unquote(backend).__gettext__(:default_domain)
+      end
+    end
+  end
+
+  defp expand_domain(domain, _env) do
+    domain
+  end
```
whatyouhide, line 238: This is pretty nasty uh? Basically if you're extracting then expand the domain at compile-time by calling the backend, and that creates a compile-time dependency but not one we care about (we're recompiling everything). If not extracting, then it's dynamic and shouldn't create a compile-time dependency.

test/gettext/new_backend_setup_test.exs
```diff
@@ -0,0 +1,62 @@
...
+        )
+
+      refute function_exported?(mod, :__gettext__, 1)
+
+      assert mod.translate() == "Hello world"
+    end
+  end
+
+  describe "compile-time dependencies" do
+    @tag :skip
```
whatyouhide, line 37: If I don't skip this, I get these errors and I have no clue why, trying to figure it out:

<details>
<summary>Output</summary>

```

  1) test allowed_locales ignores other locales as strings (GettextTest)
     test/gettext_test.exs:184
     Assertion with == failed
     code:  assert TranslatorWithAllowedLocalesString.lgettext("es", "default", nil, "Hello world", %{}) ==
              {:ok, "Hola mundo"}
     left:  {:default, "Hello world"}
     right: {:ok, "Hola mundo"}
     stacktrace:
       test/gettext_test.exs:190: (test)

........

  2) test allowed_locales ignores other locales as atom (GettextTest)
     test/gettext_test.exs:194
     Assertion with == failed
     code:  assert TranslatorWithAllowedLocalesAtom.lgettext("es", "default", nil, "Hello world", %{}) ==
              {:ok, "Hola mundo"}
     left:  {:default, "Hello world"}
     right: {:ok, "Hola mundo"}
     stacktrace:
       test/gettext_test.exs:200: (test)

.......

  3) test using a custom Gettext.Plural module with the context parameter (GettextTest)
     test/gettext_test.exs:214
     Assertion failed, no matching message after 0ms
     The process mailbox is empty.
     code: assert_received {:plural_context, %{plural_forms_header: "nplurals=2; plural=(n != 1);"}}
     stacktrace:
       test/gettext_test.exs:219: (test)

.....

  4) test known_locales/1: returns all the locales for which a backend has PO files (GettextTest)
     test/gettext_test.exs:846
     Assertion with == failed
     code:  assert Gettext.known_locales(TranslatorWithAllowedLocalesAtom) == ["es"]
     left:  []
     right: ["es"]
     stacktrace:
       test/gettext_test.exs:848: (test)

......

  5) test using a custom Gettext.Plural module (GettextTest)
     test/gettext_test.exs:204
     Assertion with == failed
     code:  assert T.lngettext("it", "default", nil, "One new email", "%{count} new emails", 1, %{}) ==
              {:ok, "1 nuove email"}
     left:  {:default, "One new email"}
     right: {:ok, "1 nuove email"}
     stacktrace:
       test/gettext_test.exs:207: (test)

.............

  6) test a custom default_domain can be set for a backend (GettextTest)
     test/gettext_test.exs:177
     Assertion with == failed
     code:  assert T.gettext("Invalid email address") == "Indirizzo email non valido"
     left:  "Invalid email address"
     right: "Indirizzo email non valido"
     stacktrace:
       test/gettext_test.exs:180: (test)
```

</details>

test/support/mix_project_helpers.ex
```diff
@@ -0,0 +1,31 @@
+defmodule GettextTest.MixProjectHelpers do
```
whatyouhide, line 1: Unnecessary refactor if it wasn't for that stupid test that is causing me issues

lib/gettext/macros.ex
```diff
@@ -0,0 +1,239 @@
+defmodule Gettext.Macros do
```
whatyouhide, line 1: José wants to put these in `Gettext.Backend` but oh man it's already confusing as it is... Would love to have these here and just make the moduledoc false 😄

josevalim: Doing it in `Gettext.Backend` is better. Today, all of the functions we invoke on the backend are defined as `@macrocallback`. By moving them to `Gettext.Backend`, we can get rid of the module callback and document the macros directly, which will be much more intuitive to developers. The callbacks should only be the actual callbacks (handle_*).

lib/gettext.ex
```diff
@@ -627,19 +627,36 @@ defmodule Gettext do
...
+      if Macro.quoted_literal?(opts) do
+        Macro.prewalk(opts, &expand_alias(&1, __CALLER__))
       else
-        # TODO: Deprecate this branch
-        require Gettext.Backend
-        Gettext.Backend.__using__(opts)
+        opts
       end
+
+    case Keyword.fetch(opts, :backend) do
```
josevalim, line 637: `opts` may be not be keyword at compile-time. So you need to check if `Keyword.keyword?(opts)` returns true.

lib/gettext/macros.ex
```diff
@@ -0,0 +1,239 @@
...
+        unquote(msgid)
+      )
+    end
+  end
+
+  defmacro dpngettext_noop(domain, msgctxt, msgid, msgid_plural) do
+    domain =
+      case domain do
+        {backend, :__default_domain__} when not is_nil(backend) ->
+          Macro.expand(backend, __CALLER__).__gettext__(:default_domain)
```
josevalim, line 53: We cannot do this. This is doing the expansion at compile-time, which means we have a compile-time dependency.

lib/gettext/macros.ex
maennchen: I think we should not have anything regarding the extraction in the backend.

So I'm in favor of having it separate. Not sure about the name `Gettext.Macros` though, I can't think of a good one right now, but it should imply that this is the consumer side of gettext and not the provider one.

lib/gettext/macros.ex
```diff
@@ -0,0 +1,239 @@
...
+        unquote(msgctxt),
+        msgid,
+        unquote(bindings)
+      )
+    end
+  end
+
+  defmacro dgettext(domain, msgid, bindings \\ Macro.escape(%{})) do
+    quote do
+      unquote(__MODULE__).dpgettext(unquote(domain), nil, unquote(msgid), unquote(bindings))
```
josevalim, line 131: I would avoid calling macros recursively, is there any chance we can move the shared logic to private functions called by macros?

lib/gettext/macros.ex
josevalim: Is the domain a string? If the domain is a string, we can support a special domain called `:default` and we resolve it at **runtime**, inside the `Gettext` module.

lib/gettext/macros.ex
josevalim: Ignore me, I think I misunderstood this.

lib/gettext/macros.ex
```diff
@@ -0,0 +1,239 @@
+defmodule Gettext.Macros do
+  @moduledoc false
```
josevalim, line 2: We should make this public and move the `@macrocallback` from `Gettext.Backend` to this module. Only the actual callbacks (`handle_*`) should remain in `Gettext.Backend`. We should also update `Gettext` module docs to say that, when it is used, it imports the macros in this module.

Maybe we should also have `use Gettext.Macros, backend: ...` instead of `use Gettext, backend: ...`. Thoughts?

maennchen: I think `use Gettext, backend: ...` is quite ergonomic and easy to remember.

I for sure don't like it with the name *Macros*, even though I can't think of a better name right now.

maennchen (commented):
We might actually have a larger issue with our current plan.

How does this work if we're not in a module context? If I for example want to use translations in an elixir script?

The only solution I can think of right now would be to have the user create two modules in separate files. A `Backend` and a `Frontend`. The frontend can then just be imported directly to invoke the gettext functions.

lib/gettext/macros.ex
josevalim: Macros make sense if you consider that you have `Gettext.dngettext` and we call them "gettext functions". It should also answer your question below: if you can't use the macros, you can always use the functions directly and pass the gettext backend as an explicit argument.

maennchen: Macros makes sense from an implementation standpoint. The user doesn’t care however that they are macros, he just wants to get the translation. If we use functions or macros is not something he cares about.

True on calling the functions directly. The import we had before was nicer, but that’s an edge case anyways and shouldn’t dictate our approach if there’s a workaround.

josevalim: I am convinced. We can keep the primary API `use Gettext, backend: …`. We need to find a name for the module and document it, but this way the name does not play such a big role anyway.

lib/gettext/macros.ex
whatyouhide: Oh wait this is a place where I forgot to do

```elixir
domain = expand_domain(domain, __CALLER__)
```

which does pretty much what you said 😄

lib/gettext/macros.ex
whatyouhide: @josevalim you ok if I do that in a subsequent PR? Already going crazy with this one and this works fine so maybe we make it work make it beautiful and stuff? 😄

whatyouhide (commented):
@josevalim @maennchen this is ready for another review.

Changes:

  * `Gettext.Macros` is public now, and hosts all the macros (documented).
  * No more `@macrocallback`s in `Gettext.Backend`. We won't generate macros anywhere now so no point in making a behaviour out of them right?
  * We use `:default` as the default domain and then expand that in `Gettext.Extractor` (at compile time) *and* in `Gettext.Macros` (at runtime, so no compile-time dependency).
  * `Gettext.Backend` introduced also the `l(n)gettext` callbacks now, why not. Good to document and stuff?

lib/gettext/macros.ex
josevalim: Yup, glad to postpone.

josevalim (approved):
Beautiful. I think the only things missing overall are:

1. A bit more docs on the `Gettext` module
2. Deprecating the old backend style
3. Test if this will actually remove compile time dependencies in a Phoenix project. You can use `mix xref trace` in a view or controller for that
4. Update Phoenix generators

:shipit:

whatyouhide:
Yes docs coming up, and deprecation as well. I tested on a small sample app (not Phoenix) and it did but I will test once I update the generatorssss

whatyouhide:
In a Phoenix project with some tweaks (the one we'll do to the generator):

```
→ mix xref trace lib/gtest_web/components/core_components.ex 
lib/gtest_web/components/core_components.ex:1: call GtestWeb.CoreComponents.__phoenix_component_verify__/1 (compile)
lib/gtest_web/components/core_components.ex:18: alias GtestWeb.Gettext (runtime)
lib/gtest_web/components/core_components.ex:568: alias GtestWeb.Gettext (runtime)
lib/gtest_web/components/core_components.ex:570: alias GtestWeb.Gettext (runtime)

→ mix xref trace lib/gtest_web/controllers/page_controller.ex
lib/gtest_web/controllers/page_controller.ex:1: call GtestWeb.PageController.__phoenix_verify_routes__/1 (compile)
lib/gtest_web/controllers/page_controller.ex:2: require GtestWeb (export)
lib/gtest_web/controllers/page_controller.ex:2: alias GtestWeb.Endpoint (runtime)
lib/gtest_web/controllers/page_controller.ex:2: alias GtestWeb.Gettext (runtime)
lib/gtest_web/controllers/page_controller.ex:2: alias GtestWeb.Layouts (runtime)
lib/gtest_web/controllers/page_controller.ex:2: alias GtestWeb.Router (runtime)
lib/gtest_web/controllers/page_controller.ex:2: call GtestWeb.__using__/1 (compile)
lib/gtest_web/controllers/page_controller.ex:2: call GtestWeb.static_paths/0 (compile)
```

Looks fantastique!

PR #392, opened 2024-08-18
Deprecate "use Gettext" for backends and improve docs

whatyouhide:


lib/gettext.ex
```diff
@@ -642,8 +615,23 @@ defmodule Gettext do
...
+
+        is deprecated. To define a backend, call:
+
+            use Gettext.Backend, otp_app: :my_app
+
+        Then, to use the backend, call this in your module:
+
+            use Gettext, backend: MyApp.Gettext
+
+        """)
```
josevalim, line 632:
```suggestion
        """, __CALLER__)
```

Or something similar.

josevalim, line 632:
```suggestion
        """, Macro.Env.from_stacktrace(__CALLER__))
```

From Elixir v1.14, you can also just pass `__CALLER__`.

lib/gettext.ex
```diff
@@ -642,8 +615,23 @@ defmodule Gettext do
...
+
         quote do
-          # TODO: Deprecate this branch
           use Gettext.Backend, unquote(opts)
```
josevalim, line 635: I have an idea. Currently, `use Gettext` defines a bunch of macros. However, we don't want `use Gettext.Backend` to define those macros, as they should use the new `use Gettext, backend: ...` API. So what if we change this code to be:

```
quote do
  use Gettext.Backend, unquote(opts)
  unquote(macros())
end
```

where those are deprecated macros?

lib/gettext.ex
whatyouhide: Nice!

lib/gettext.ex
Gladear: fwiw, I think that's a good idea, it will prevent `import MyApp.Gettext` from working on updated installations, which will help with consistency and maintenance.

whatyouhide:
@josevalim done with the `_with_backend` stuff we talked about. There's refactoring to do but this works and I would love to merge this pass 😄

lib/gettext/compiler.ex
```diff
@@ -99,12 +97,12 @@ defmodule Gettext.Compiler do
     end)
   end
 
-  defp macros() do
+  def generate_macros(_env) do
```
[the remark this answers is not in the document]
whatyouhide: Yeah anyway it's going away so wherever it lives it won't live there for long hihi

lib/gettext.ex
whatyouhide: Done!

lib/gettext/macros.ex
```diff
@@ -394,16 +396,351 @@ defmodule Gettext.Macros do
...
+        unquote(msgid_plural)
+      )
+    end
+  end
+
+  @doc """
+  Same as `pngettext_noop/3`, but takes an explicit backend.
+  """
+  defmacro pngettext_noop_with_backend(backend, msgctxt, msgid, msgid_plural) do
+    backend = Gettext.Compiler.expand_backend(backend, __CALLER__)
```
josevalim, line 522: Honestly, I would not try to share this code, and instead have two implementations of `expand_backend`, one for macros and another for the compiler. Especially because we want to remove the one in compiler anyway.

josevalim (approved):

PR #393, opened 2024-08-19
Refactor Gettext.Macros to generate less code

whatyouhide:
Talked about this with @josevalim this morning.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -33,7 +33,7 @@ defmodule Gettext.Mixfile do
           "Locale Functions": &(&1[:section] == :locale),
 
           # Gettext.Macros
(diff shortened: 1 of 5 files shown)


mix.exs
```diff
@@ -33,8 +33,11 @@ defmodule Gettext.Mixfile do
           "Locale Functions": &(&1[:section] == :locale),
 
           # Gettext.Macros
+          "Macros with Explicit Backend":
```
josevalim, line 36: I think this will be too long for the sidebar. Perhaps "With Backend" is enough?

josevalim (approved):

PR #398, opened 2024-09-10
Use Expo Metadata Stripping to improve Performance

maennchen:


lib/gettext/compiler.ex
```diff
@@ -499,7 +499,9 @@ defmodule Gettext.Compiler do
   # lngettext/7 (for plural messages) clauses.
   defp compile_po_file(kind, po_file, env, plural_mod, interpolation_module) do
     %{locale: locale, domain: domain, path: path} = po_file
-    %Messages{messages: messages, file: file} = messages_struct = PO.parse_file!(path)
+
+    %Messages{messages: messages, file: file} =
+      messages_struct = PO.parse_file!(path, strip_meta: true)
```
whatyouhide, line 504: Won't this break unless people use Expo 1.1? We can check for Expo 1.1 at compile time with `Application.spec(:vsn)`?

maennchen: We don't validate for unknown options. I have verified that it doesn't fail with expo 0.5.1.

whatyouhide: Okay great.

whatyouhide (approved):
🎉

PR #401, opened 2024-11-12
Warn user if conflicting plural messages are defined

### LINKED ISSUE #400: No error on conflicting plural messages
Hi,

i dont't know if this is really a `gettext` issue but when we cache the `_build` folder in our GitlLab CI Pipeline the command `mix gettext.extract --check-up-to-date` although it runs perfectly fine locally.
I suspect some `mtime` issues because the cache is restored after the repo is cloned. 

Any ideas on how to prevent this?

<details>
<summary>The Job definition</summary>

```yaml
locales:
  stage: lint
  image: $ELIXIR_IMAGE
  cache:
    - key: $CI_COMMIT_REF_SLUG
      paths:
        - _build
    - key:
        files:
          - mix.lock
      paths:
        - deps
  before_script:
    - apk add --no-cache make gcc libc-dev git
    - mix local.rebar --force
    - mix local.hex --force
    - mix deps.get
  script:
    - mix gettext.extract --check-up-to-date
  interruptible: true
  allow_failure: true
```

</details>

maennchen:
Fixes #400

### DIFF
--- a/lib/gettext/extractor_agent.ex
+++ b/lib/gettext/extractor_agent.ex
@@ -3,6 +3,8 @@ defmodule Gettext.ExtractorAgent do
 
   use Agent
 
+  require Logger
+
   alias Expo.Message
 
   @name __MODULE__
@@ -75,12 +77,45 @@ defmodule Gettext.ExtractorAgent do
     end)
   end
 
-  defp merge_messages(message_1, message_2) do
+  defp merge_messages(%Message.Singular{} = message_1, %Message.Plural{} = message_2) do
+    # Flipping the arguments to make sure that the pluaral message (more information) is used as the base message
+    merge_messages(message_2, message_1)
+  end
+
+  defp merge_messages(%Message.Plural{} = message_1, %Message.Plural{} = message_2) do
+    # Make sure message choice is deterministic
+    [message_1, message_2] =
+      Enum.sort_by([message_1, message_2], &IO.iodata_to_binary(&1.msgid_plural))
+
+    if IO.iodata_to_binary(message_1.msgid_plural) != IO.iodata_to_binary(message_2.msgid_plural) do
+      Logger.warning("""
+      Plural message for '#{IO.iodata_to_binary(message_1.msgid)}' is not matching:
+      Using '#{IO.iodata_to_binary(message_2.msgid_plural)}' instead of '#{IO.iodata_to_binary(message_1.msgid_plural)}'.
+      References: #{dump_references(message_1.references ++ message_2.references)}\
+      """)
+    end
+
+    merge_messages_after_checks(message_1, message_2)
+  end
+
+  defp merge_messages(message_1, message_2), do: merge_messages_after_checks(message_1, message_2)
+
+  defp merge_messages_after_checks(message_1, message_2) do
     message_1
     |> Map.put(:references, message_1.references ++ message_2.references)
     |> Map.put(
       :extracted_comments,
       Enum.uniq(message_1.extracted_comments ++ message_2.extracted_comments)
     )
+  end
+
+  defp dump_references(references) do
+    references
+    |> List.flatten()
+    |> Enum.map(fn
+      {file, line} -> [file, ":", Integer.to_string(line)]
+      file -> file
+    end)
+    |> Enum.intersperse(", ")
   end
 end
--- a/test/gettext/extractor_test.exs
+++ b/test/gettext/extractor_test.exs
@@ -1,6 +1,8 @@
 defmodule Gettext.ExtractorTest do
   use ExUnit.Case
 
+  import ExUnit.CaptureLog
+
   alias Expo.Message
   alias Expo.Messages
   alias Gettext.Extractor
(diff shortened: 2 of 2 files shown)


lib/gettext/extractor_agent.ex
```diff
@@ -75,7 +77,26 @@ defmodule Gettext.ExtractorAgent do
     end)
   end
 
-  defp merge_messages(message_1, message_2) do
+  defp merge_messages(message_1, message_2)
+
```
whatyouhide, line 81: Generally I don't think we do bodiless clauses for private functions. We can shave off a couple of lines here 🙃 

```suggestion
```

lib/gettext/extractor_agent.ex
```diff
@@ -75,7 +77,26 @@ defmodule Gettext.ExtractorAgent do
...
+  defp merge_messages(%Message.Singular{} = message_1, %Message.Plural{} = message_2) do
+    # Flipping the arguments to make sure that the pluaral message (more information) is used as the base message
+    merge_messages(message_2, message_1)
+  end
+
+  defp merge_messages(%Message.Plural{} = message_1, %Message.Plural{} = message_2) do
+    if IO.iodata_to_binary(message_1.msgid_plural) != IO.iodata_to_binary(message_2.msgid_plural) do
+      Logger.warning(
+        "Plural message for '#{IO.iodata_to_binary(message_1.msgid)}' is not matching. Using '#{IO.iodata_to_binary(message_2.msgid_plural)}' instead of '#{IO.iodata_to_binary(message_1.msgid_plural)}'."
+      )
```
whatyouhide, line 91: Is this enough information for someone to find the problematic message in case they have a larger set of PO(T) files?

lib/gettext/extractor_agent.ex
```diff
@@ -75,7 +77,26 @@ defmodule Gettext.ExtractorAgent do
...
+    end
+
+    _merge_messages(message_1, message_2)
+  end
+
+  defp merge_messages(message_1, message_2), do: _merge_messages(message_1, message_2)
+
+  defp _merge_messages(message_1, message_2) do
```
whatyouhide, line 99: Nit: we can call this `merge_messages_after_checks` or something to avoid the somewhat-uncommon `_` prefix.

whatyouhide (commented):
Fantastic work.

lib/gettext/extractor_agent.ex
maennchen: How do you like the new message?

```
Plural message for '{msgid}' is not matching:
Using '{new plural msgid}' instead of '{old plural msgid}'.
References: [{file:line}]
```

lib/gettext/extractor_agent.ex
```diff
@@ -75,12 +77,45 @@ defmodule Gettext.ExtractorAgent do
...
-  defp merge_messages(message_1, message_2) do
+  defp merge_messages(%Message.Singular{} = message_1, %Message.Plural{} = message_2) do
+    # Flipping the arguments to make sure that the pluaral message (more information) is used as the base message
+    merge_messages(message_2, message_1)
+  end
+
+  defp merge_messages(%Message.Plural{} = message_1, %Message.Plural{} = message_2) do
+    # Make sure message choice is deterministic
+    [message_1, message_2] =
+      Enum.sort_by([message_1, message_2], &IO.iodata_to_binary(&1.msgid_plural))
```
maennchen, line 88: @whatyouhide I also added this instruction to make sure that we get a deterministic result no matter the compilation order when spanning multiple files.

lib/gettext/extractor_agent.ex
```diff
@@ -75,12 +77,45 @@ defmodule Gettext.ExtractorAgent do
...
+
+    if IO.iodata_to_binary(message_1.msgid_plural) != IO.iodata_to_binary(message_2.msgid_plural) do
+      Logger.warning("""
+      Plural message for '#{IO.iodata_to_binary(message_1.msgid)}' is not matching:
+      Using '#{IO.iodata_to_binary(message_2.msgid_plural)}' instead of '#{IO.iodata_to_binary(message_1.msgid_plural)}'.
+      References: #{dump_references(message_1.references ++ message_2.references)}\
+      """)
```
whatyouhide, line 95: Should we use `IO.warn/2` instead? Do we use Logger in other places? I can't recall

whatyouhide (approved):
Looks great. If we use Logger in other places too then this is good to go.

lib/gettext/extractor_agent.ex
maennchen: We use Logger for:

* Missing bindings - https://github.com/elixir-gettext/gettext/blob/b41445ea04da137c62fb7ebc822fd8f05f2353fa/lib/gettext/backend.ex#L45
* Domain contains slashes - https://github.com/elixir-gettext/gettext/blob/b41445ea04da137c62fb7ebc822fd8f05f2353fa/lib/gettext/compiler.ex#L383
* Missing Plural Forms - https://github.com/elixir-gettext/gettext/blob/b41445ea04da137c62fb7ebc822fd8f05f2353fa/lib/gettext/compiler.ex#L723

whatyouhide: Nice, sounds good then! Go for it.

PR #402, opened 2024-11-13
Release v0.27.0

maennchen:
I would consider the metadata stripping functionality a new feature and would therefore go to 0.27.0 instead of 0.26.2.

But since this does not directly affect the user, one could also argue for a patch release. What do you prefer?


whatyouhide:
Let's go with patch because pre-1.0 most folks would likely have `~> 0.26.0` and wouldn't catch this update.

PR #403, opened 2024-11-13
Add legalese

maennchen:
I just saw that the project did not include the full license. (besides the hint in the README)

I though this would be a good time to add some legalese to the project:

* LICENSE - Apache 2.0
* CODE_OF_CONDUCT - [Contributor Covenant](https://www.contributor-covenant.org/)
* SECURITY - Copied & adapted from EEF
* CONTRIBUTION - Copied & adapted from EEF

## TODO

* [ ] Enable private vulnerability reporting in project
* [ ] Set correct email for disclosures in SECURITY - which one?
* [ ] Set correct email for abuse in CODE_OF_CONDUCT - which one?


whatyouhide:
- I’m ok with the license (great actually!) and CONTRIBUTING.
- Code of conduct is good, but can you pick it from https://github.com/elixir-lang/elixir/blob/main/CODE_OF_CONDUCT.md or Phoenix? We don't really follow EEF conventions.
- I don't think we need `SECURITY.md` in a project like Gettext.

Thanks!

maennchen:
I can adapt the elixir code of conduct for sure. Do you want to copy it and just adapt the scope & email address or would you just like to refer to it?

While I agree that we will probably not have security issues with this project, I would still like to include a security policy. The reason for this is mainly that there’s tool out there like the OpenSSF Scorecard project, which is used by corporates to check their dependencies. Having a security policy defined is a good sign for compliance departments and will also result in higher scores in those tools. If this was a small and barely used library, I probably wouldn’t bother. But with the popularity, I think it’s worth to set it up.

maennchen:
Btw: Phoenix also uses the Contributors Covenant, just an older version. While the EEF uses it as well, that’s not the reason I chose it. I believe it’s one of the most prevalent codes of conduct. Elixir is also based on it.

whatyouhide:
> I can adapt the elixir code of conduct for sure. Do you want to copy it and just adapt the scope & email address or would you just like to refer to it?

Referring is great for now, we don't need a specific one here. There is no active "community" around Gettext.

Ok, let's go with the security doc too.

maennchen:
@whatyouhide I changed the reference to the elixir code of conduct.

For the security report email I would offer to receive the emails myself.

(In case I would get something that actually matters I would then just create an entry with GitHub Vulnerability Reporting and coordinate with all maintainers from there.)

PR #412, opened 2025-04-11
Fix concatenation of sigils

### LINKED ISSUE #411: `gettext/1` and related macro do not support `~S` concatenated strings
Hello 👋 

I like to split my translations over multiple lines to avoid them being too long. I also like to use the `~S` sigil when there are `"` in the string to avoid escaping them. It appears both are not compatible currently 😄 

```elixir
gettext(
  "a multi-line" <>
    ~S(string with sigil)
)
```

This results in a compilation error: `** (ArgumentError) Gettext macros expect message keys (msgid and msgid_plural), domains, and comments to expand to strings at compile-time, but the given msgid doesn't.`

I would expect this to work, especially since `gettext(~S("quoted string"))` works fine 👌

Gladear:
Add support for concatenation with sigils (e.g. `"Hello " <> ~s(world)`).

Fixes #411

### DIFF
--- a/lib/gettext/macros.ex
+++ b/lib/gettext/macros.ex
@@ -709,25 +709,25 @@ defmodule Gettext.Macros do
       """
     end
 
-    do_expand_to_binary(term, env, raiser)
+    validated_expand_to_binary(term, env, raiser)
   end
 
-  defp do_expand_to_binary({:<>, _, pieces}, env, raiser) do
-    Enum.map_join(pieces, &do_expand_to_binary(&1, env, raiser))
+  defp validated_expand_to_binary({:<>, _, pieces}, env, raiser) do
+    Enum.map_join(pieces, &validated_expand_to_binary(&1, env, raiser))
   end
 
-  defp do_expand_to_binary({:<<>>, _, pieces}, env, raiser) do
-    Enum.map_join(pieces, &do_expand_to_binary(&1, env, raiser))
+  defp validated_expand_to_binary({:<<>>, _, pieces}, env, raiser) do
+    Enum.map_join(pieces, &validated_expand_to_binary(&1, env, raiser))
   end
 
   # We support nil too in order to fall back to a nil context and always use the *p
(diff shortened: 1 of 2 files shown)


test/gettext/macros_test.exs
```diff
@@ -15,12 +15,19 @@ defmodule Gettext.MacrosTest do
...
       assert gettext("Hello world") == "Ciao mondo"
       assert gettext(@gettext_msgid) == "Ciao mondo"
       assert gettext(~s(Hello world)) == "Ciao mondo"
+      assert gettext("Hello " <> "world") == "Ciao mondo"
+      assert gettext("Hello " <> "wor" <> "ld") == "Ciao mondo"
+      assert gettext(<<"Hello world">>) == "Ciao mondo"
+      assert gettext(<<"Hello ", "world">>) == "Ciao mondo"
+      assert gettext("Hello " <> <<"wor", "ld">>) == "Ciao mondo"
+      assert gettext("Hello " <> ~s(world)) == "Ciao mondo"
+      assert gettext(~S(Hello ) <> ~s(world)) == "Ciao mondo"
```
whatyouhide, line 30: Fantastic thank you!

lib/gettext/macros.ex
```diff
@@ -709,15 +709,29 @@ defmodule Gettext.Macros do
       """
     end
 
-    # We support nil too in order to fall back to a nil context and always use the *p
-    # variants of the Gettext macros.
+    do_expand_to_binary(term, env, raiser)
```
whatyouhide, line 712: We kind of moved away from the convention of `do_` prefixes because of their "imperative" feel. Instead, here what if we call this
```suggestion
    validated_expand_to_binary(term, env, raiser)
```

?

whatyouhide (commented):
Just one comment and we're ready to ship!

lib/gettext/macros.ex
Gladear: OK ! Changed 👌

PR #417, opened 2025-07-25
Don't time out when running the extract task

### LINKED ISSUE #416: Timeout running "mix gettext.extract --merge"
Hi! Thanks for this library!

We have an application with a fairly large messages to be translated, and a pending PR added a large chunk of new calls to `gettext`. When running `mix gettext.extract --merge`, I see:

```
Compiling 567 files (.ex)
Generating PortalCore.Cldr for 7 locales named [:de, :en, :es, :fr, :it, ...] with a default locale named :en
Compiling lib/portal/postgrex_types.ex (it's taking more than 10s)
Compiling lib/portal_core/cldr.ex (it's taking more than 10s)
Generated portal app
Wrote priv/gettext/en/LC_MESSAGES/errors.po (0 new messages, 0 removed, 24 unchanged, 0 reworded (fuzzy), 0 marked as obsolete)
** (exit) exited in: Task.Supervised.stream(5000)
    ** (EXIT) time out
    (elixir 1.18.4) lib/task/supervised.ex:345: Task.Supervised.stream_reduce/7
    (elixir 1.18.4) lib/stream.ex:703: Stream.run/1
    (gettext 0.26.2) lib/mix/tasks/gettext.extract.ex:85: Mix.Tasks.Gettext.Extract.run_message_extraction/3
    (mix 1.18.4) lib/mix/task.ex:495: anonymous fn/3 in Mix.Task.run_task/5
    (mix 1.18.4) lib/mix/cli.ex:107: Mix.CLI.run_task/2
```

This is fixed locally by adding adding `timeout: :infinity` to the call to `Task.stream_async` in `lib/mix/tasks/gettext.merge.ex`. I will add a PR, but wanted to reference an issue.

jhosteny:
Set a timeout of :infinity when running extract. This is in line with the setting for the merge operation.

Fixes #416

### DIFF
--- a/lib/mix/tasks/gettext.extract.ex
+++ b/lib/mix/tasks/gettext.extract.ex
@@ -129,7 +129,7 @@ defmodule Mix.Tasks.Gettext.Extract do
     pot_files
     |> Enum.map(fn {path, _} -> Path.dirname(path) end)
     |> Enum.uniq()
-    |> Task.async_stream(&Mix.Tasks.Gettext.Merge.run([&1 | argv]), ordered: false)
+    |> Task.async_stream(&Mix.Tasks.Gettext.Merge.run([&1 | argv]), ordered: false, timeout: :infinity)
     |> Stream.run()
   end
 end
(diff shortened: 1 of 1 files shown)


whatyouhide (approved):

jhosteny:
Hi @whatyouhide, sure! I pushed another commit with that change. Note that that are two other uses of `:infinity` for task timeouts. I didn't change those, since I am not sure if a timeout of 120s is currently ever being exceeded for those. In this case, merging this would relax the timeout from 5s to 120s, so it seems safer.

PR #419, opened 2025-08-24
Add new locale function put_locale_protected/2

abdelhalim97:
## Changes 
Adding new function `put_locale_protected/2` inspired from  `put_locale/2` only difference that when unsupported locale is given it will fallback to english `en` like [default locale](https://hexdocs.pm/gettext/Gettext.html#module-default-locale)

## The why
I m trying to add this function because few months ago we had issue where we expected the `put_locale/2` to not change the language when not supported language has been given(problem caused mainly from our side but still) i had to fork the repo and do changes in our organization, but now i said why dont i try to make them public to everyone

## Questions
Also i thought about changing the original `put_locale/2` instead of creating new function but its not good idea no?
What do you think about  `put_locale_protected/1` without backend
Changing the fallback from just english to be taken from the `default_locale`

___________
Why no unit test? I m not sure if the feature will get approved so i didnt want to invest much time
This is my first ever contribution to open source, i m open to all suggestions :pray:

### DIFF
--- a/lib/gettext.ex
+++ b/lib/gettext.ex
@@ -772,6 +772,29 @@ defmodule Gettext do
     do: raise(ArgumentError, "put_locale/2 only accepts binary locales, got: #{inspect(locale)}")
 
   @doc """
+  Similar to `put_locale/2`, but it checks if the passed locale exists in the known_locales or not.
+  If yes it sets the locale for the current process and the given `backend`.
+  Else it falls back to the default locale for the backend.
+  ## Examples
+
+      Gettext.known_locales(MyApp.Gettext)
+      #=> ["en", "ja"]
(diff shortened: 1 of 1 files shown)


abdelhalim97:
@maennchen Sorry for tagging you but i have no reviewers assigned so probobly no one got notif about it
Could you check the pr please

abdelhalim97:
Thanks maennchen for your reply, i didnt want to rush you but just to make sure you know about the PR
_______

@whatyouhide thanks for the review

> will not work for everyone

Got it

> improve the documentation for put_locale/2
> We could also consider put_locale!/2

I love put_locale!/2 idea i can work on that, soon i ll close this one and open new Pr with the suggested changes

PR #420, opened 2025-08-31
Add new function put_locale!/2 

abdelhalim97:
## Changes 

- Enhancing documentation for `put_locale/2`
- Add new function `put_locale!/2` inspired from `put_locale/2` only difference that if unsupported locale is introduced it wil raise error


lib/gettext.ex
```diff
@@ -1154,4 +1178,18 @@ defmodule Gettext do
...
   defp domain_or_default(_backend, domain) when is_binary(domain), do: domain
+
+  @spec put_locale_with_fallback(backend, locale) :: binary() | nil
+  defp put_locale_with_fallback(backend, locale) do
+    allowed_locales = known_locales(backend)
+    is_allowed_locale = locale in allowed_locales
+
+    case is_allowed_locale do
+      true ->
+        Process.put(backend, locale)
```
maennchen, line 1189: Can you call `put_locale(backend, locale)` here? Just in case we ever change the mechanism.

maennchen:
@abdelhalim97 Thanks for the PR :heart: 

@whatyouhide Besides the small nit above, this looks like a good improvement.

abdelhalim97:
Done :saluting_face: 
Thanks for your review

maennchen (approved):

PR #421, opened 2025-09-09
 Solve problem with erasing .po after recompile

PapaHusserl:
I've got a few issues with re-enabling "extract". Some of the .po partly erased. I pretty sure that removing only manifests and cleaning elixir's compile file isn't enough. So I suggest to clean the whole app_path to be ensure that nothing could interrupt extracting. 
I need some time to be perfectly sure it will work, so I leave PR as WIP until that.


josevalim:
This is fixed in gettext v1.0.2 but will require upcoming Elixir v1.19.3, thanks!

PR #423, opened 2025-10-28
Removed unnecessary cleaning of Elixir manifests

### LINKED ISSUE #373: Duplicate references in pot files and warnings about redefining modules.
When I run 
`mix gettext.extract --merge`

on my umrella project, I get a LOT of warnings saying:

```
warning: redefining module Core.Reports.ReportThree.Report (current version loaded from /_build/dev/lib/core/ebin/Elixir.Core.Reports.ReportThree.Report.beam)
  lib/core/reports/report_three.ex:24
```
  
Even if I delete `_build` - I think this is also causing this issue, on all of my POT & PO files, I get

```
#: lib/core/reports/report_three.ex:170
#: lib/core/reports/report_three.ex:170
#, elixir-autogen, elixir-format
msgid "Error"
msgstr "Error"
```

Which I assume is caused by the module being in memory twice or something, duplicating the reference.
I now just `uniq` the generated PO and POT files as a workaround, but I'd rather understand what the core issue is here.

josevalim:
Closes #373.

### DIFF
--- a/lib/mix/tasks/gettext.extract.ex
+++ b/lib/mix/tasks/gettext.extract.ex
@@ -112,9 +112,6 @@ defmodule Mix.Tasks.Gettext.Extract do
   end
 
   defp force_compile do
-    Mix.Tasks.Compile.Elixir.clean()
-    Enum.each(Mix.Tasks.Compile.Elixir.manifests(), &File.rm/1)
-
     # If "compile" was never called, the reenabling is a no-op and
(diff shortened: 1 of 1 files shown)


Fudoshiki:
mix gettxet.extract --merge and mix gettext.merge priv/gettext dosn't update locales now

josevalim:
@Fudoshiki you are right. I need to add a feature that forces only the Elixir compiler to run. I will revert and ship a new version.

PR #433, opened 2026-05-26
Fix Elixir 1.20 compilation warnings

gilbertwong96:
Resolve compilation warnings in Elixir 1.20.


lib/gettext/compiler.ex
```diff
@@ -423,7 +423,7 @@ defmodule Gettext.Compiler do
           :parallel ->
             grouped
             |> Enum.map(fn {module, files} ->
-              Kernel.ParallelCompiler.async(fn ->
+              Task.async(fn ->
```
josevalim: It must use the Kernel.ParallelCompiler.pmap. Using Task is wrong as it cannot track modules.

josevalim: You will also need to bump the minimum Elixir version on mix.exs and CI.

.tool-versions
josevalim, line 1: Please remove this file :)

lib/gettext/extractor.ex
```diff
@@ -388,11 +388,6 @@ defmodule Gettext.Extractor do
     end
   end
 
-  defp ensure_empty_msgstr!(%Message.Plural{} = message) do
-    raise Error,
-          "plural message with msgid '#{IO.iodata_to_binary(message.msgid)}' has a non-empty msgstr"
-  end
-
```
whatyouhide, line 395: This was dead code, correct?

gilbertwong96: yes, the previous function clause already covers all case and compiler reports warning for this clause.

josevalim (approved):

whatyouhide:
Thanks @gilbertwong96 💟

PR #441, opened 2026-07-29
Deduplicate source references when merging extracted messages

rigzad:
Two extraction calls that resolve to the same msgid at the same file:line each contribute a reference group, and merge_messages_after_checks/2 concatenates them without uniquing. Macro-generated code produces N calls attributed to a single line, so the same `#:` comment is written N times.

`extracted_comments` on the very next line is already uniqued; this makes `references` consistent with it.

### DIFF
--- a/test/gettext/extractor_test.exs
+++ b/test/gettext/extractor_test.exs
@@ -476,6 +476,41 @@ defmodule Gettext.ExtractorTest do
     Extractor.disable()
   end
 
+  test "duplicate references for one message are recorded once" do
+    Extractor.enable()
+
+    code = """
+    defmodule Gettext.ExtractorTest.SameLineGettext do
+    use Gettext.Backend, otp_app: :test_application
+    end
+
+    defmodule SameLine do
+    require Gettext.Macros
+
+    def bar do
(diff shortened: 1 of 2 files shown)


maennchen (approved):

maennchen:
@rigzad Thanks for the PR ❤️
Can you run `mix format` please?

rigzad:
@maennchen done. My bad !

test/gettext/extractor_test.exs
```diff
@@ -476,6 +476,41 @@ defmodule Gettext.ExtractorTest do
...
+    use Gettext.Backend, otp_app: :test_application
+    end
+
+    defmodule SameLine do
+    require Gettext.Macros
+
+    def bar do
+      [Gettext.Macros.gettext_with_backend(Gettext.ExtractorTest.SameLineGettext, "foo"), Gettext.Macros.gettext_with_backend(Gettext.ExtractorTest.SameLineGettext, "foo")]
+    end
+    end
```
maennchen, line 493:
```suggestion
    defmodule Gettext.ExtractorTest.SameLineGettext do
      use Gettext.Backend, otp_app: :test_application
    end

    defmodule SameLine do
      require Gettext.Macros

      def bar do
        [Gettext.Macros.gettext_with_backend(Gettext.ExtractorTest.SameLineGettext, "foo"), Gettext.Macros.gettext_with_backend(Gettext.ExtractorTest.SameLineGettext, "foo")]
      end
    end
```

